PR #2, opened 2022-02-23
fix colors

### LINKED ISSUE #1: Colors seems to be inverted
Hi !

I tested this new implementation [here](https://github.com/pyaillet/ttgo-display-idf-rs/blob/8709f33701ef3283fa444455eba01dbd2d5f18e2/src/main.rs). I faced a strange bug, it seems colors are inverted (See picture below).

I will have a look a make a PR if I can identify the problem and a solution.

IMG_20220223_205821

pyaillet:
Fix #1

Not sure if this works for every setup but I got it working [there](https://github.com/pyaillet/ttgo-display-idf-rs/commit/1494245b1406fd50f6dfb0197859a14d464893bc)


almindor:
Hmm this is odd as I tested it on my ST7789 and it worked without the invert. I'll double check, give me a day or two to get the setup back to working order :)

pyaillet:
If you don't mind I rather close this one and open another PR with the full init code taken from the original ST7789 driver.

Issue #3, opened 2022-02-24
Optional RST handling


pyaillet:
Hi !

I was able to use this driver for a TTGO Display with st7789. However there is no RST pin on the TTGO T-Watch.
I would like to propose some changes to this driver to handle the absence of the RST Pin.

Would you prefer handling this by wrapping the RST Pin in an Option ? Creating different models ? Another approach ?

almindor:
Good question. I wonder if this should be handled by the `DisplayInterface` instead. That'd be a big change tho, for now I think we should make it optional somehow.

The issue is using just `Option<RST>` doesn't solve it because it still requires the type to be `OutputPin` of some kind. So even if you're passing `None` you'd have to specify the type. We could possibly add a constructor that uses some sort of dummy `OutputPin` type internally so users can avoid that issue I guess.

pyaillet:
Let's close this one two as the PR was merged :)

Thanks again !

PR #4, opened 2022-02-25
add the support for displays without a RST pin

pyaillet:
The choice was made here to rely on an internal NoPin implementation to hide the details
and expose it as an Alias type.

Proposal fix for #3 

It would be used like this : 
```rust
let display = DisplayNoRST::st7789_without_rst(di);
```

And it would be defined as:
```rust
  pub struct Twatch<'a> {
      pmu: Pmu<'a>,
      display: DisplayNoRST<EspSpi2InterfaceNoCS, mipidsi::models::ST7789>,
      [...]
  }
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,6 +31,7 @@ pub mod models;
 use models::Model;
 
 mod graphics;
+mod no_pin;
 
 #[cfg(feature = "batch")]
 mod batch;
@@ -53,6 +54,8 @@ where
     // Current orientation
     orientation: Orientation,
 }
+
+pub type DisplayNoRST<DI, MODEL> = Display<DI, no_pin::NoPin, MODEL>;
 
 ///
 /// Display orientation.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,6 +32,7 @@ use models::Model;
 
 mod graphics;
 mod no_pin;
+pub use no_pin::*;
 
 #[cfg(feature = "batch")]
 mod batch;
@@ -48,14 +49,12 @@ where
     // Display interface
     di: DI,
     // Reset pin.
-    rst: RST,
+    rst: Option<RST>,
     // Model
     model: MODEL,
     // Current orientation
     orientation: Orientation,
 }
-
-pub type DisplayNoRST<DI, MODEL> = Display<DI, no_pin::NoPin, MODEL>;
 
 ///
 /// Display orientation.
@@ -118,7 +117,7 @@ where
     /// * `rst` - display hard reset [OutputPin]
     /// * `model` - the display [Model]
     ///
-    pub fn with_model(di: DI, rst: RST, model: M) -> Self {
+    pub fn with_model(di: DI, rst: Option<RST>, model: M) -> Self {
         Self {
             di,
             rst,
@@ -220,7 +219,7 @@ where
     /// Release resources allocated to this driver back.
     /// This returns the display interface and the RST pin deconstructing the driver.
     ///
-    pub fn release(self) -> (DI, RST, M) {
+    pub fn release(self) -> (DI, Option<RST>, M) {
         (self.di, self.rst, self.model)
     }
 
--- a/src/models/st7789.rs
+++ b/src/models/st7789.rs
@@ -98,7 +98,7 @@ where
     DI: WriteOnlyDataCommand,
 {
     ///
-    /// Creates a new [Display] instance with [ST7789] as the [Model] and without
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -54,6 +55,8 @@ where
     orientation: Orientation,
 }
 
+pub type DisplayNoRST<DI, MODEL> = Display<DI, no_pin::NoPin, MODEL>;
```
almindor, line 58: Do we need another type tho? I'm thinking making `rst: Option<RST>` with the additional `no_rst` contstructor variants will be less confusing.

almindor: It'll also force model implementors to account for this,  which atm. the ST7789 doesn't for example (if it works it works by accident. I think I should do a soft reset in that case).

pyaillet: Do you mean the `NoPin` type or the `DisplayNoRST` type ?
1. I am using a custom NoPin type to avoid users to choose a random Pin when they mean to pass a None value for the RST Pin.
2. I am defining a DisplayNoRST type to avoid leaking the `NoPin` type externally

I don't know how it would be possible to use Option<RST> without needing to provide a specific Pin as a driver user, but any suggestion is welcome.

almindor: I mean 2.

Leaking `NoPin` is inevitable in this case, but at least the user won't really care, they just do `Display::st7789_no_rst(di)` or such and that's it from their perspective.

pyaillet: I tried to rewrite according to your remarks.

However you will see in the commit 48ed7cababb5d2da2a08ca4f7a47dea5aa6cf9ae that I took the init sequence from your old driver, as it solved the problem with the inverted colors for my TTGO T-Watch-2020.

src/no_pin.rs
```diff
@@ -0,0 +1,21 @@
+use embedded_hal::digital::v2::{OutputPin, PinState};
+
+use core::convert::Infallible;
+
+#[derive(Default)]
+pub struct NoPin {}
```
almindor, line 6:
```suggestion
pub struct NoPin;
```

src/no_pin.rs
```diff
@@ -0,0 +1,21 @@
+use embedded_hal::digital::v2::{OutputPin, PinState};
+
+use core::convert::Infallible;
+
+#[derive(Default)]
```
almindor, line 5: could you add a basic documentation point here? just explain it's a dummy for the no-cs case.

src/no_pin.rs
```diff
@@ -0,0 +1,21 @@
...
+pub struct NoPin {}
+
+impl OutputPin for NoPin {
+    type Error = Infallible;
+
+    fn set_low(&mut self) -> Result<(), Self::Error> {
+        Ok(())
+    }
```
almindor, line 13:
```suggestion
    }
    
```

src/models/st7789.rs
```diff
@@ -84,6 +93,24 @@ where
...
+        Self::with_model(di, Some(rst), ST7789::new())
+    }
+}
+
+impl<DI> Display<DI, NoPin, ST7789>
+where
+    DI: WriteOnlyDataCommand,
+{
+    ///
+    /// Creates a new [Display] instance with [ST7789] as the [Model] and without
```
almindor, line 105:
```suggestion
    /// Creates a new [Display] instance with [ST7789] as the [Model] without
```

src/models/st7789.rs
```diff
@@ -20,24 +21,31 @@ impl Model for ST7789 {
...
+            Some(ref mut rst) => self.hard_reset(rst, delay)?,
+            None => write_command(di, Instruction::SWRESET, &[])?,
+        }
+        delay.delay_us(150_000);
 
         write_command(di, Instruction::SLPOUT, &[])?; // turn off sleep
+        delay.delay_us(10_000);
 
-        write_command(di, Instruction::COLMOD, &[0b0101_0101])?; // 16bit 65k colors
+        write_command(di, Instruction::INVOFF, &[])?;
```
almindor, line 40: let's keep this to a separate PR so we don't mix logical changes together.

almindor (changes requested):
Great job! thanks. Just a few nits + let's keep the init code change into a separate PR so we don't mix things too much.

pyaillet:
Many thanks for the review :)

I removed the init seq from this PR and most of the comments shoulds be resolved now.

almindor (approved):
Thanks, this is great! I'll make a release after we get the init code in and tested out.

PR #9, opened 2022-03-20
Fix size and orientation issues

### LINKED ISSUE #7: screen orientations are hardcoded
Some screens require different madctl 0x26 settings, and e.g. will appear reversed without.

For a possible solution see: https://github.com/yuri91/ili9341-rs/commit/8a4aee95c2934cd6aaac4b243ab8bda24d0b7475

### LINKED ISSUE #8: screen size hardcoded and does not change with orientation
Screen size is hard-coded and should vary depending on the orientation settings.

brianmay:
Fixes #7
Fixes #8

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -67,10 +67,10 @@ where
     }
 
     fn fill_solid(&mut self, area: &Rectangle, color: Self::Color) -> Result<(), Self::Error> {
-        let fb_size = self.model.framebuffer_size();
+        let fb_size = self.size();
         let fb_rect = Rectangle::with_corners(
             Point::new(0, 0),
-            Point::new(fb_size.0 as i32, fb_size.1 as i32),
+            Point::new(fb_size.width as i32, fb_size.height as i32),
         );
         let area = area.intersection(&fb_rect);
 
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -95,18 +95,16 @@ where
     }
 
     fn clear(&mut self, color: Self::Color) -> Result<(), Self::Error>
-    where
-        Self: Sized,
     {
         let fb_size = self.model.framebuffer_size();
         let pixel_count = usize::from(fb_size.0) * usize::from(fb_size.1);
         let colors = core::iter::repeat(color).take(pixel_count); // blank entire HW RAM contents
 
         match self.orientation {
-            Orientation::Portrait | Orientation::PortraitSwapped => {
+            Orientation::Portrait => {
                 self.set_pixels(0, 0, fb_size.0, fb_size.1, colors)
             }
-            Orientation::Landscape | Orientation::LandscapeSwapped => {
+            Orientation::Landscape => {
                 self.set_pixels(0, 0, fb_size.1, fb_size.0, colors)
             }
         }
@@ -121,6 +119,10 @@ where
 {
     fn size(&self) -> Size {
         let ds = self.model.display_size();
-        Size::new(u32::from(ds.0), u32::from(ds.1))
+        let (width, height) = match self.orientation {
+            Orientation::Portrait => (ds.0, ds.1),
+            Orientation::Landscape => (ds.1, ds.0)
+        };
+        Size::new(u32::from(width), u32::from(height))
     }
 }
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -67,10 +67,10 @@ where
     }
 
     fn fill_solid(&mut self, area: &Rectangle, color: Self::Color) -> Result<(), Self::Error> {
-        let fb_size = self.size();
+        let fb_size = self.framebuffer_size();
         let fb_rect = Rectangle::with_corners(
             Point::new(0, 0),
-            Point::new(fb_size.width as i32, fb_size.height as i32),
+            Point::new(fb_size.0 as i32, fb_size.1 as i32),
         );
         let area = area.intersection(&fb_rect);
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -64,8 +64,6 @@ where
 pub enum Orientation {
     Portrait = 0b0000_0000,         // no inverting
     Landscape = 0b0110_0000,        // invert column and page/column order
```
brianmay, line 66: I would also be setting this to just `0b0010_0000` as the 2nd bit can be inverted if required.

almindor:
Sorry about the delay, I need to get my setup in working order again after losing an mcu to a short. I should be able to test these in a week or so.

brianmay:
Rebased to latest master.

src/graphics.rs
```diff
@@ -95,18 +95,16 @@ where
     }
 
     fn clear(&mut self, color: Self::Color) -> Result<(), Self::Error>
-    where
-        Self: Sized,
```
brianmay, line 99: I don't remember making this change. I don't see any good reason for the including these lines, so I left it in.

src/lib.rs
almindor: I'm wondering if it wouldn't be more straightforward to have a flat enum for all the possible combinations rather than the "invert x/y".

src/graphics.rs
almindor: I think Self: Sized is implied either way?

src/graphics.rs
```diff
@@ -67,10 +67,10 @@ where
     }
 
     fn fill_solid(&mut self, area: &Rectangle, color: Self::Color) -> Result<(), Self::Error> {
-        let fb_size = self.model.framebuffer_size();
```
almindor, line 70: I'm still conflicted about this. Framebuffer_size is supposed to be "the whole pixel memory" which in case of clear is what we want AFAICS.

src/graphics.rs
brianmay: Yes, I think it is.

src/graphics.rs
brianmay: Clear doesn't use `fill_solid`. At least not with this version. So nothing here is going to affect the clear function.

However I do have an alternative approach. Will add another commit, see if you like that any better.

src/lib.rs
brianmay: That is going to be a lot of combinations. Basically there are 3 bits, so 2*2*2 = 8 combinations. Think it might be more confusing, not less confusing.

Although personally, wondering if it would be better just to have:

```
pub enum Orientation {
    Portrait,
    Landscape,
}
```

And have the `set_orientation` function map that to the correct value.

src/graphics.rs
brianmay: I added a `framebuffer_size` function and use that now, hopefully this will address your concern here.

brianmay: Having said that, there is still the problem that both "whole pixel memory" and "displayable area" both need to start at 0,0. I don't think the concept is very usable as is.

src/graphics.rs
```diff
@@ -8,6 +8,19 @@ use crate::models::Model;
 use crate::{Display, Error, Orientation};
 use display_interface::WriteOnlyDataCommand;
 
+fn framebuffer_size<DI, RST, M>(display: &Display<DI, RST, M>) -> (u16, u16)
```
brianmay, line 11: Thought about making this return a `Size` object. That would clarify which one is width vs height. But Size objects use u32 dimensions. And `pixel_count` below uses `usize` dimensions. And `usize::from` works fine for u16, but fails for u32. So I stuck with the above for now.

src/lib.rs
```diff
@@ -242,6 +251,19 @@ where
             .map_err(|_| Error::DisplayError)
     }
 
+    fn framebuffer_size(&self) -> (u16, u16)
```
brianmay, line 254: Thought about making this return a `Size` object. That would clarify which one is width vs height. But Size objects use u32 dimensions. And `pixel_count` in `graphics.rs` uses `usize` dimensions. And `usize::from` works fine for u16, but fails for u32. So I stuck with the above for now.

src/lib.rs
brianmay: Also worth noting, yes, we really do need to support all of 8 combinations. Depending on how the display is mounted and if the hardware reverses the image or not.

almindor: Hmm the main issue I have with this is that splitting the orientation enum and the "revert" values seem confusing. We could do something like
```rust
pub enum Orientation {
    Portrait(bool, bool), // x/y invert
    Landscape(bool, bool), // x/y invert
}
```

Or possibly introduce a new sub-enum for "inversion" and use that as the inner value.

almindor:
@brianmay what is instead of having "two world" (model vs display/lib) we unify the `framebuffer_size` in the model by either enforcing the orientation knowledge in the model, or by making it into `fn frabebuffer_size(&self, orientation: Orientation)` (provided `Orientation` has all the required info as I'd like).

This way we can keep a single source of size info. Same can be done for `display_size` then.

src/lib.rs
brianmay: Hmmm. Not sure I can see how this could be less confusing. Also most of the code only needs to know if it is Portrait/Landscape. If the display is inverted or not makes no difference. i.e. the invert parameters are only really needed to configure the MADCTL register and don't matter afterward. In fact the invert parameters have no real meaning outside the MADCTL because there is no real way to know what the hardware is doing with them. Maybe setting invert_x will invert the display, or maybe setting invert_x will restore the display.

Maybe that is the problem though. We use the `set_orientation` to set parameters that are not orientation. You wouldn't set the RGB/BGR value here. But the problem being it is interconnected. When setting orientation you typically will (always) have to adjust the invert values at the same time. Which is why we do it in the `set_orientation` call.

brianmay:
> @brianmay what is instead of having "two world" (model vs display/lib) we unify the `framebuffer_size` in the model by either enforcing the orientation knowledge in the model, or by making it into `fn frabebuffer_size(&self, orientation: Orientation)` (provided `Orientation` has all the required info as I'd like).
> 
> This way we can keep a single source of size info. Same can be done for `display_size` then.

I would be happy with both of these approaches. At present the model has no state information, the display is the authoritative source for all information. We probably should keep this unless we have a good reason to change it (e.g. if RGB/BGR config requires it). If so, then the later option would suffice. Also note that we only need to know the Portrait/Landscape stuff, we don't need to know bout the invert stuff here.

src/lib.rs
almindor: `u32 as usize` should be safe in 32bit+ machines of course there are very special edge cases (there's a whole thing with ptrsize vs usize in rust). for our purposes and expected display sizes I think it's 100% safe to just case that way.

almindor (approved):
I'm going to merge this and then follow up with a PR to clarify with the orientation being pushed into `Model::*_size()` functions instead.

PR #12, opened 2022-03-27
Add call to `init` in README.md

### LINKED ISSUE #6: common_init function never called
When calling this as per instructions:

```rust
// create a DisplayInterface from SPI and DC pin, with no manual CS control
let di = SPIInterfaceNoCS::new(spi, dc);
// create the ILI9486 display driver in rgb666 color mode from the display interface and RST pin
let mut display = Display::ili9486_rgb666(di, rst);
// clear the display to black
display.clear(Rgb666::BLACK)?;
```

The `common_init` function is never called, and the screen is not initialised.

brianmay:
Fixes #6.


almindor (approved):
Thank you!

Issue #10, opened 2022-03-23
red/blue swapped


brianmay:
~This is kind of weird:~

* ~white is correct~
* ~red is correct~
* ~blue is correct~
* ~green is correct~
* ~magenta comes correct (red + blue)~
* ~yellow comes out correct (red + green) ~
* blue comes out correct from arduino app

~But with this library:~

* ~blue comes out as red.~
* ~cyan comes out as yellow (green + blue).~

LATE: Sorry, disregard most of the above. I tested again, red comes out as blue and blue comes out as red. It does work correctly with arduino.

This seems to be a clear case of red and blue being swapped. But green remains the same.

It is consistent. Filling screen fills it entirely with the same colour. Happens for all draw operations.

With this change:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 8026834..9e1d54e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ display-interface = "0.4.1"
 embedded-graphics-core = "0.3.3"
 embedded-hal = "0.2.6"
 nb = "1.0.0"
+log = "0.4"
 
 [dependencies.heapless]
 optional = true
diff --git a/src/models/ili9486.rs b/src/models/ili9486.rs
index 4c1475b..cbdb2ae 100644
--- a/src/models/ili9486.rs
+++ b/src/models/ili9486.rs
@@ -96,9 +96,11 @@ impl Model for ILI9486Rgb666 {
     {
         write_command(di, Instruction::RAMWR, &[])?;
         let mut iter = colors.into_iter().flat_map(|c| {
+            use log::*;
             let red = c.r() << 2;
             let green = c.g() << 2;
             let blue = c.b() << 2;
+            error!("ddd {red:?} {green:?} {blue:?}");
             [red, green, blue]
         });
 ```
 
 I get a lot of:
 
 ```
 1970-01-01T00:06:11.158Z ERROR mipidsi::models::ili9486                > ddd 0 0 252
 ```

This looks like the correct BLUE colour to me. But this was displayed as RED.

So no idea what is going on here. But it does that something weird is going on with the colours.

So maybe the hardware has swapped the red and blue lines? But strange the arduino app works.

brianmay:
Swapping red/blue in `/src/models/ili9486.rs` fixes the problem. But still weird. As Green is OK unchanged, and the standard clearly states that the colour order is red, green, and blue.

[...]

PR #13, opened 2022-03-27
Switched to Bgr to fix Red/Blue colours being swapped

KerryRJ:
Fixes [#10](https://github.com/almindor/mipidsi/issues/10) for the st7735s.

### DIFF
--- a/src/models/st7735s.rs
+++ b/src/models/st7735s.rs
@@ -30,14 +30,15 @@ impl Model for ST7735s {
     {
         match rst {
             Some(ref mut rst) => self.hard_reset(rst, delay)?,
-            None => write_command(di, Instruction::SWRESET, &[])?,
+            None => {
+                write_command(di, Instruction::SWRESET, &[])?;
+                delay.delay_us(120);
+            }
         }
-        delay.delay_us(200_000);
 
         write_command(di, Instruction::SLPOUT, &[])?; // turn off sleep
         delay.delay_us(120_000);
 
-        write_command(di, Instruction::INVON, &[])?; // turn inversion on
         write_command(di, Instruction::INVON, &[])?; // turn inversion on
         write_command(di, Instruction::FRMCTR1, &[0x05, 0x3A, 0x3A])?; // set frame rate
         write_command(di, Instruction::FRMCTR2, &[0x05, 0x3A, 0x3A])?; // set frame rate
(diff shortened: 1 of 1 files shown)


src/models/st7735s.rs
```diff
@@ -52,7 +52,7 @@ impl Model for ST7735s {
         write_command(di, Instruction::PGC, &[0x10, 0x0E, 0x02, 0x03, 0x0E, 0x07, 0x02, 0x07, 0x0A, 0x12, 0x27, 0x37, 0x00, 0x0D, 0x0E, 0x10])?; // set GAMMA +Polarity characteristics
         write_command(di, Instruction::NGC, &[0x10, 0x0E, 0x03, 0x03, 0x0F, 0x06, 0x02, 0x08, 0x0A, 0x13, 0x26, 0x36, 0x00, 0x0D, 0x0E, 0x10])?;  // set GAMMA -Polarity characteristics
         write_command(di, Instruction::COLMOD, &[0b0101_0101])?; // set interface pixel format, 16bit pixel into frame memory
-        write_command(di, Instruction::MADCTL, &[0b1010_1000])?; // set memory data access control, Top -> Bottom, RGB, Left -> Right
+        write_command(di, Instruction::MADCTL, &[0b1010_0000])?; // set memory data access control, Top -> Bottom, RGB, Left -> Right
```
brianmay, line 55: This bit changed is MH, "Horizontal refresh order control". Probably don't need to change this in order to fix the colours.

brianmay (changes requested):

brianmay (changes requested):
Main problem with this is that while it will fix BGR systems, it will break RGB systems I think. Needs to be configurable somehow.

LATER: Then again, if you know that if you are sure that all st7735s devices use BGR, it should be fine...

src/models/st7735s.rs
KerryRJ: I just double checked [Datasheet 1.4](https://github.com/michal037/driver-ST7735S/blob/master/documentation/datasheet/st7735s-datasheet-v1.4.pdf) and D3 is RGB.

Do you have a link to the Datasheet you have available?

brianmay: Sorry, I was out by 1 bit, reading the docs. Your are right. Trying to understand the need to change the `ColorFormat`, would it be sufficient just to change the MADCTL value?

KerryRJ:
The datasheets that I have referenced stipulate RGB565.

Checking the C code showed that Big Endian was used to write the pixel data. I briefly tried using Little Endian to write the data instead. This solved the red/blue problem but then created a problem with  the 6 bits for green. Switching the MADCTL bit and the ColourFormat solves red/green/blue for this particular device.

I do not have additional st7735s based devices to verify if this will work for all iterations. Making it configurable would be a good solution. At this point in time I do not have enough experience to do this.

src/models/st7735s.rs
KerryRJ: I did try the MADCTL bit change first, but it has no impact. I then tried using Little Endian instead of Big Endian. That solves the red/blue but then swaps half of the green bits.

Just confirming whether this C code uses LE to write the data?
```C
static void LCD_0IN96_SendData_16Bit(UWORD Data)
{
    DEV_Digital_Write(LCD_DC_PIN, 1);
    DEV_Digital_Write(LCD_CS_PIN, 0);
    DEV_SPI_WriteByte((Data >> 8) & 0xFF);
    DEV_SPI_WriteByte(Data);
    DEV_Digital_Write(LCD_CS_PIN, 1);
}
```

KerryRJ: Setting the MADCTL bit had no effect without also switching to BGR.

Just confirming that this is indeed BE C code?
```C
static void LCD_0IN96_SendData_16Bit(UWORD Data)
{
    DEV_Digital_Write(LCD_DC_PIN, 1);
    DEV_Digital_Write(LCD_CS_PIN, 0);
    DEV_SPI_WriteByte((Data >> 8) & 0xFF);
    DEV_SPI_WriteByte(Data);
    DEV_Digital_Write(LCD_CS_PIN, 1);
}
```

brianmay: I might be getting confused here. To me it sounds like setting the MADCTL RGB bit has no effect. I just tried setting this bit on my ILI99488 device, which supposedly has the same MADCTL RGB bit. Absolutely no change that I can see, all colours are the exactly same. Odd. Wonder what the point of this config option is when it does nothing. Maybe hardware dependant.

I think you swap the colours by switching the colour model, and that means when you call `c.into_storage()` the colours get serialised in the alternate order:

https://github.com/almindor/mipidsi/blob/master/src/models/st7789.rs#L63

I believe there might be two ways to make this configurable:

1. At compile time. Make `ColorFormat` a generic parameter to the `ST7789` struct. I *think* this should be possible...

2. Do the serialisation "by hand". e.g. have a look at https://github.com/almindor/mipidsi/blob/master/src/models/ili9486.rs#L98-L103 - ideally for efficiency probably only want to do this in such a way that you test config once, not once for every individual colour.

KerryRJ: I have switched to serialising manually after finding this in the C drivers clear method

```C
    Color = ((Color<<8)&0xff00)|(Color>>8);
```

although this just swaps the MSB and the LSB I found that the G range of colours would be wrong doing it this way.

PR #14, opened 2022-03-31
keep MADCTL value to use with orientation changes

### LINKED ISSUE #7: screen orientations are hardcoded

almindor:
Keeps MADCTL value from model `init` and uses it with orientation writes to keep the color and other definitions.

Should fix #7 and #8 and enable a fix for #10

I've also switched to pure static dispatch for all cases with `WriteOnlyDataCommand`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,6 +54,8 @@ where
     model: MODEL,
     // Current orientation
     orientation: Orientation,
+    // Current MADCTL value
+    madctl: u8,
 }
 
 ///
@@ -123,6 +125,7 @@ where
             rst,
             model,
             orientation: Orientation::default(),
+            madctl: 0,
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -148,9 +152,11 @@ where
     /// Sets display [Orientation]
     ///
     pub fn set_orientation(&mut self, orientation: Orientation) -> Result<(), Error<RST::Error>> {
+        let value = (self.madctl & 0b0001_1111) | (orientation as u8);
```
brianmay, line 155: Am wondering if there is anything we should do to make it crystal clear that the orientation value should never be setting the lower 5 bits, Maybe: `orientation & 0b1110_0000` or `orientation << 5`?

brianmay (approved):
I agree with this change and think this is the correct strategy for fixing the problem that updating the orientation will overwrite the MADCTL values.

brianmay:
I see nothing here that will fix #7 or #8. And fix for #10 still requires some way of setting the madctl value which looks like it is still hard coded per each driver. But good step in correct direction.

src/lib.rs
almindor: good point

almindor:
> I see nothing here that will fix #7 or #8. And fix for #10 still requires some way of setting the madctl value which looks like it is still hard coded per each driver. But good step in correct direction.

Correct. I'll check #9 for the orientation fix once it's rebased.

I'll make a new change for exposing the color inversion once we confirm it works with this fixed.

PR #16, opened 2022-04-05
make Model size functions orientation aware

almindor:
Makes `framebuffer_size` and `display_size` methods `Orientation` aware to simplify the logic.

@brianmay @KerryRJ - please review if possible, :+1: if you think it's good to go.


brianmay (approved):

brianmay:
I haven't tested it by running it, however everything looks good to me.

PR #17, opened 2022-04-06
use DisplayOptions to allow MADCTL control, fixes #10

### LINKED ISSUE #10: red/blue swapped

almindor:
Introduces `DisplayOptions` with all possible `MADCTL` values abstracted into a more digestible struct.

I've also refactored `Orientation` to include inversion logic so we don't have `init` vs `set_orientation` asymmetry.

This should fix #10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,6 +211,7 @@ where
         self.madctl = self
             .model
             .init(&mut self.di, &mut self.rst, delay_source, options)?;
+        self.orientation = options.orientation;
         Ok(())
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,6 +108,21 @@ pub enum TearingEffect {
 }
 
 ///
+/// Defines expected color component ordering, RGB or BGR
+///
+#[derive(Copy, Clone)]
+pub enum ColorOrder {
+    Rgb,
+    Bgr,
+}
+
+impl Default for ColorOrder {
+    fn default() -> Self {
+        Self::Rgb
+    }
+}
+
+///
 /// Options for displays used on initialization
 ///
 #[derive(Copy, Clone, Default)]
@@ -116,8 +131,8 @@ pub struct DisplayOptions {
     pub orientation: Orientation,
     /// Set to make display vertical refresh bottom to top
     pub invert_vertical_refresh: bool,
-    /// Set to make display use BGR instead of RGB pixel format
-    pub rgb_to_bgr: bool,
+    /// Specify display color ordering
+    pub color_order: ColorOrder,
     /// Set to make display horizontal refresh right to left
     pub invert_horizontal_refresh: bool,
 }
@@ -129,8 +144,9 @@ impl DisplayOptions {
         if self.invert_vertical_refresh {
             value |= 0b0001_0000;
         }
-        if self.rgb_to_bgr {
-            value |= 0b0000_1000;
+        match self.color_order {
+            ColorOrder::Rgb => {}
+            ColorOrder::Bgr => value |= 0b0000_1000,
         }
         if self.invert_horizontal_refresh {
             value |= 0b0000_0100;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,7 +86,7 @@ impl Orientation {
 
     /// Shortcut function to create un-inverted Landscape Orientation
     pub fn landscape() -> Self {
-        Self::Portrait(Inversion::None)
+        Self::Landscape(Inversion::None)
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,54 +59,37 @@ where
 }
 
 ///
-/// Display orientation inversion.
-///
-#[derive(Copy, Clone)]
-pub enum Inversion {
-    None,
-    X,
-    Y,
-    XY,
-}
-
-///
 /// Display orientation.
 ///
 #[derive(Copy, Clone)]
 pub enum Orientation {
-    Portrait(Inversion),
-    Landscape(Inversion),
-}
-
-impl Orientation {
-    /// Shortcut function to create un-inverted Portrait Orientation
-    pub fn portrait() -> Self {
-        Self::Portrait(Inversion::None)
-    }
-
-    /// Shortcut function to create un-inverted Landscape Orientation
-    pub fn landscape() -> Self {
-        Self::Landscape(Inversion::None)
-    }
+    /// Portrait orientation, with mirror image parameter
+    Portrait(bool),
+    /// Landscape orientation, with mirror image parameter
+    Landscape(bool),
+    /// Inverted Portrait orientation, with mirror image parameter
+    PortraitInverted(bool),
+    /// Inverted Lanscape orientation, with mirror image parameter
+    LandscapeInverted(bool),
 }
 
 impl Default for Orientation {
     fn default() -> Self {
-        Self::Portrait(Inversion::None)
+        Self::Portrait(false)
     }
 }
 
 impl Orientation {
     pub fn value_u8(&self) -> u8 {
         match self {
-            Orientation::Portrait(Inversion::None) => 0b0000_0000,
-            Orientation::Portrait(Inversion::X) => 0b1000_0000,
-            Orientation::Portrait(Inversion::Y) => 0b0100_0000,
-            Orientation::Portrait(Inversion::XY) => 0b1100_0000,
-            Orientation::Landscape(Inversion::None) => 0b0110_0000,
-            Orientation::Landscape(Inversion::X) => 0b0010_0000,
-            Orientation::Landscape(Inversion::Y) => 0b1110_0000,
-            Orientation::Landscape(Inversion::XY) => 0b1010_0000,
+            Orientation::Portrait(false) => 0b0000_0000,
+            Orientation::Portrait(true) => 0b0100_0000,
+            Orientation::PortraitInverted(false) => 0b1100_0000,
+            Orientation::PortraitInverted(true) => 0b1000_0000,
+            Orientation::Landscape(false) => 0b0010_0000,
+            Orientation::Landscape(true) => 0b0110_0000,
+            Orientation::LandscapeInverted(false) => 0b1110_0000,
+            Orientation::LandscapeInverted(true) => 0b1010_0000,
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,18 +59,55 @@ where
 }
 
 ///
+/// Display orientation inversion.
+///
+#[derive(Copy, Clone)]
+pub enum Inversion {
+    None,
+    X,
+    Y,
+    XY,
+}
+
+///
 /// Display orientation.
 ///
-#[repr(u8)]
 #[derive(Copy, Clone)]
 pub enum Orientation {
-    Portrait = 0b0000_0000,  // no inverting
-    Landscape = 0b0110_0000, // invert column and page/column order
+    Portrait(Inversion),
+    Landscape(Inversion),
+}
+
+impl Orientation {
+    /// Shortcut function to create un-inverted Portrait Orientation
+    pub fn portrait() -> Self {
+        Self::Portrait(Inversion::None)
+    }
+
+    /// Shortcut function to create un-inverted Landscape Orientation
+    pub fn landscape() -> Self {
+        Self::Portrait(Inversion::None)
+    }
 }
 
 impl Default for Orientation {
     fn default() -> Self {
-        Self::Portrait
+        Self::Portrait(Inversion::None)
+    }
+}
+
+impl Orientation {
+    pub fn value_u8(&self) -> u8 {
+        match self {
+            Orientation::Portrait(Inversion::None) => 0b0000_0000,
+            Orientation::Portrait(Inversion::X) => 0b1000_0000,
+            Orientation::Portrait(Inversion::Y) => 0b0100_0000,
+            Orientation::Portrait(Inversion::XY) => 0b1100_0000,
+            Orientation::Landscape(Inversion::None) => 0b0110_0000,
+            Orientation::Landscape(Inversion::X) => 0b0010_0000,
+            Orientation::Landscape(Inversion::Y) => 0b1110_0000,
+            Orientation::Landscape(Inversion::XY) => 0b1010_0000,
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


almindor:
@brianmay @KerryRJ happy to hear your thoughts on this

src/lib.rs
```diff
@@ -87,6 +124,39 @@ pub enum TearingEffect {
...
+/// Options for displays used on initialization
+///
+#[derive(Copy, Clone, Default)]
+pub struct DisplayOptions {
+    /// Initial display orientation (without inverts)
+    pub orientation: Orientation,
+    /// Set to make display vertical refresh bottom to top
+    pub invert_vertical_refresh: bool,
+    /// Set to make display use BGR instead of RGB pixel format
+    pub rgb_to_bgr: bool,
```
KerryRJ, line 137: This doesn't fit the st7735s as it needs to be true but then it sounds like we're using BGR when it is supposed to be RGB. I am assuming that the st7789 is the same.

almindor:
> I am finding that switching between the portrait()/landscape() methods, and the enums(Inversion) a bit strange. Might just be me. I would prefer if @brianmay weighs in on this.

I agree it's not ideal. The problem is if we disconnect the inverts then `set_orientation` becomes asymmetric with `init` which is silly, basically preventing anyone from initializing the display inverted "just because".

I'm open to suggestions that'd make this more ergonomic given this constraint.

src/lib.rs
```diff
@@ -58,19 +58,56 @@ where
...
+
+impl Orientation {
+    /// Shortcut function to create un-inverted Portrait Orientation
+    pub fn portrait() -> Self {
+        Self::Portrait(Inversion::None)
+    }
+
+    /// Shortcut function to create un-inverted Landscape Orientation
+    pub fn landscape() -> Self {
+        Self::Portrait(Inversion::None)
```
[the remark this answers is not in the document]
almindor: fixed, thanks!

src/lib.rs
brianmay: I suggest instead of a bool we use an `Enum { BGB, RGB}` here.

brianmay:
Taking a step back, and doing a complete rethink, I believe that there are two use cases for setting the inversion bits (actually MY and MX I think):

1. My screen is mirror image! I need to hold it to a mirror just to read the writing! Toggling one of the invert lines is required. Contrary to what I though previously, It doesn't actually matter which one, if you get the wrong one, the screen will appear upside-down. In which case the next use case will apply.

2. My screen is upside down! Toggle *both* of the invert lines from the values previously set.

So how about:

1, For the init function we have an argument `mirror_image: bool`.  This will set one of the invert bits. Either one. But not both.
2. The orientations enum values have an invert boolean. i.e.:

```rust
pub enum Orientation {
    Portrait(bool),
    Landscape(bool),
}

Or maybe we just have 4 variations like before. The bool flag means "use this orientation but display upside-down".

If this bool flag is false, we do nothing.  If the bool flag is true, then we toggle *both* of MY and MX. If the values were both 0, we set them to both 1. If one of the values was set previously in step 1, then we set that bit to 0, and the other bit to 1.

The theory is that once the `mirror_image` bit is set correctly, you should be able to use all 4 possible combinations for `Orientation`, and all of them will display correctly without being reversed.

Assumption: I believe that toggling the MV (rotate) bit should not cause the image to be reversed, on any display. At least this is the case for my display. Correct me if I am wrong. If I am wrong, might need to have two `mirror_portrait` and `mirror_landscape` parameters that only take affect for the respective orientation. But hoping this isn't required.

**Comment by KerryRJ** [comment]
If memory serves me correct the st7735s needs both invert bits set to swap the image other wise the text ends up reversed on a single plane. I'd need to double check to confirm.

**Comment by brianmay** [comment]
Yes, this is as I said. You toggle one bit to reverse the display, or toggle both bits to turn the display upside down.

**Comment by almindor** [comment]
> Taking a step back, and doing a complete rethink, I believe that there are two use cases for setting the inversion bits (actually MY and MX I think):
> 
> 1. My screen is mirror image! I need to hold it to a mirror just to read the writing! Toggling one of the invert lines is required. Contrary to what I though previously, It doesn't actually matter which one, if you get the wrong one, the screen will appear upside-down. In which case the next use case will apply.
> 2. My screen is upside down! Toggle _both_ of the invert lines from the values previously set.
> 
> So how about:
> 
> 1, For the init function we have an argument `mirror_image: bool`. This will set one of the invert bits. Either one. But not both. 2. The orientations enum values have an invert boolean. i.e.:
> 
> ```rust
> pub enum Orientation {
>     Portrait(bool),
>     Landscape(bool),
> }
> ```
> 
> Or maybe we just have 4 variations like before. The bool flag means "use this orientation but display upside-down".
> 
> If this bool flag is false, we do nothing. If the bool flag is true, then we toggle _both_ of MY and MX. If the values were both 0, we set them to both 1. If one of the values was set previously in step 1, then we set that bit to 0, and the other bit to 1.
> 
> The theory is that once the `mirror_image` bit is set correctly, you should be able to use all 4 possible combinations for `Orientation`, and all of them will display correctly without being reversed.
> 
> Assumption: I believe that toggling the MV (rotate) bit should not cause the image to be reversed, on any display. At least this is the case for my display. Correct me if I am wrong. If I am wrong, might need to have two `mirror_portrait` and `mirror_landscape` parameters that only take affect for the respective orientation. But hoping this isn't required.

Wait, if we wanted to do it as 4 enum values what would the MADCTL look like for each?

```rust
#[repr(u8)]
#[derive(Copy, Clone)]
pub enum Orientation {
    Portrait = 0b0000_0000,
    Landscape = 0b0110_0000,
    PortraitInverted = ???,
    LandscapeInverted = ???,
}
```

brianmay:
> Wait, if we wanted to do it as 4 enum values what would the MADCTL look like for each?

Depends on if mirror was set to true or not in the init call.

So for the init call:

```rust
fn init(mirror: bool)  {
    let madctl = if mirror {0x0b1000_0000 } else {  0b0000_0000 }
}
```

And for the set_orientation call:

```rust
fn set_orientation(&mut self, orientation: Orientation) {
    let madctl = match orientation {
        Portrait => 0b0000_0000 | self.madctl & 0b1101_1111,
        Landscape => 0b0010_0000 | self.madctl & 0b1101_1111,
        PortraitInverted => 0b0000_0000  | (self.madctl ^ 0x0b1100_0000) & 0b1101_1111
        LandscapeInverted => 0b0010_0000| (self.madctl ^ 0x0b1100_0000) & 0b1101_1111
    }
    self.madctl = mactl;
    // output new madctl value
}
```

Perhaps a truth table might help:

Mirror | Orientation | MX | MY | MV
-- | -- | -- | -- | --
FALSE | Portrait | FALSE | FALSE | FALSE
FALSE | Landscape | FALSE | FALSE | TRUE
FALSE | Portrait Inverse | TRUE | TRUE | FALSE
FALSE | Landscape Inverse | TRUE | TRUE | TRUE
TRUE | Portrait | TRUE | FALSE | FALSE
TRUE | Landscape | TRUE | FALSE | TRUE
TRUE | Portrait Inverse | FALSE | TRUE | FALSE
TRUE | Landscape Inverse | FALSE | TRUE | TRUE

almindor:
> > Wait, if we wanted to do it as 4 enum values what would the MADCTL look like for each?
> 
> Depends on if mirror was set to true or not in the init call.
> 
> So for the init call:
> 
> ```rust
> fn init(mirror: bool)  {
>     let madctl = if mirror {0x0b1000_0000 } else {  0b0000_0000 }
> }
> ```
> 
> And for the set_orientation call:
> 
> ```rust
> fn set_orientation(&mut self, orientation: Orientation) {
>     let madctl = match orientation {
>         Portrait => 0b0000_0000 | self.madctl & 0b1101_1111,
>         Landscape => 0b0010_0000 | self.madctl & 0b1101_1111,
>         PortraitInverted => 0b0000_0000  | (self.madctl ^ 0x0b1100_0000) & 0b1101_1111
>         LandscapeInverted => 0b0010_0000 | (self.madctl ^ 0x0b1100_0000) & 0b1101_1111
>     }
>     self.madctl = mactl;
>     // output new madctl value
> }
> ```
> 
> Perhaps a truth table might help:
> 
> Mirror	Orientation	MX	MY	MV
> FALSE	Portrait	FALSE	FALSE	FALSE
> FALSE	Landscape	FALSE	FALSE	TRUE
> FALSE	Portrait Inverse	TRUE	TRUE	FALSE
> FALSE	Landscape Inverse	TRUE	TRUE	TRUE
> TRUE	Portrait	TRUE	FALSE	FALSE
> TRUE	Landscape	TRUE	FALSE	TRUE
> TRUE	Portrait Inverse	FALSE	TRUE	FALSE
> TRUE	Landscape Inverse	FALSE	TRUE	TRUE

This means that we cannot just have 4 values in the `Orientation` enum then tho. We'd still need to have either a `(bool)` or store the `mirror` boolean somewhere else somehow.

brianmay:
> This means that we cannot just have 4 values in the Orientation enum then tho. We'd still need to have either a (bool) or store the mirror boolean somewhere else somehow.

That is one way.

You can also store the values in the madctl and use XOR operation to toggle the bits, as I have shown in the sample (untested) code above.

Possibly a boolean might make the code easier to read, not sure.

almindor:
> > This means that we cannot just have 4 values in the Orientation enum then tho. We'd still need to have either a (bool) or store the mirror boolean somewhere else somehow.
> 
> That is one way.
> 
> You can also store the values in the madctl and use XOR operation to toggle the bits, as I have shown in the sample (untested) code above.
> 
> Possibly a boolean might make the code easier to read, not sure.

Added the bool to `Orientation` + some more documentation to it.

src/lib.rs
almindor: `Model::init` should take care of the inverting the bit logic here. The display options can stay out of it, but I agree about the enum.

src/lib.rs
```diff
@@ -134,8 +203,14 @@ where
     ///
     /// * `delay_source` - mutable reference to a [DelayUs] provider
     ///
-    pub fn init(&mut self, delay_source: &mut impl DelayUs<u32>) -> Result<(), Error<RST::Error>> {
-        self.madctl = self.model.init(&mut self.di, &mut self.rst, delay_source)?;
+    pub fn init(
+        &mut self,
+        delay_source: &mut impl DelayUs<u32>,
+        options: DisplayOptions,
+    ) -> Result<(), Error<RST::Error>> {
```
brianmay, line 210: Need to have a line here that sets `self.orientation`:

```
self.orientation = options.orientation;
```

Otherwise screen dimensions come out wrong. Unless you are call `set_orientation`.

brianmay (approved):
I think it looks good to me. Just need to fix the bug when you set Landscape in the init call, but screen dimensions don't get updated.

brianmay:
Actually not sure what is going on with this example:

```rust
let options = DisplayOptions {
        orientation: Orientation::Landscape(false),
        invert_vertical_refresh: false,
        color_order: ColorOrder::Bgr,
        invert_horizontal_refresh: false,
    };
    display.init(&mut delay::Ets, options).unwrap();
    display
        .set_orientation(Orientation::Portrait(false))
        .unwrap();
```

I get Portrait screen dimensions, but the screen appears in Landscape. Huh?

OH, wonder if the problem is this line on `set_orientation`:

```
let value = self.madctl | (orientation.value_u8() & 0b1110_0000);
```

Probably need to use `self.madctl & 0b0001_1111` here.

brianmay:
Contrary to what I said above (and I thought had tested too) `Portrait(false)` is mirrored but `Landscape(false)` isn't. Weird. Using `Portrait(true)` fixes it, so not a problem. But thought I should correct what I said before.

almindor:
> Contrary to what I said above (and I thought had tested too) `Portrait(false)` is mirrored but `Landscape(false)` isn't. Weird. Using `Portrait(true)` fixes it, so not a problem. But thought I should correct what I said before.

Wait, I'm confused now. When you say `Portrait(false)` is mirrored, do you mean it's bugged and should be swapped with what `Portrait(true)` gives instead?

brianmay:
> Wait, I'm confused now. When you say Portrait(false) is mirrored, do you mean it's bugged and should be swapped with what Portrait(true) gives instead?

On my hardware maybe. This might be hardware specific, not sure, I only have one device to test. The current situation with this PR doesn't really bother me though,

brianmay:
Don't forget to get the `set_orientation`function call as per https://github.com/almindor/mipidsi/pull/17#issuecomment-1094387568

almindor:
> Actually not sure what is going on with this example:
> 
> ```rust
> let options = DisplayOptions {
>         orientation: Orientation::Landscape(false),
>         invert_vertical_refresh: false,
>         color_order: ColorOrder::Bgr,
>         invert_horizontal_refresh: false,
>     };
>     display.init(&mut delay::Ets, options).unwrap();
>     display
>         .set_orientation(Orientation::Portrait(false))
>         .unwrap();
> ```
> 
> I get Portrait screen dimensions, but the screen appears in Landscape. Huh?
> 
> OH, wonder if the problem is this line on `set_orientation`:
> 
> ```
> let value = self.madctl | (orientation.value_u8() & 0b1110_0000);
> ```
> 
> Probably need to use `self.madctl & 0b0001_1111` here.

Fixed

brianmay (approved):

PR #18, opened 2022-06-05
add scroll region setting

pyaillet:
Hi !

Here is a proposition to add the support for `Vertical Scrolling Definition`.

I was able to test and validate this on `ST7789`


almindor (approved):
Looking good thanks!

PR #25, opened 2022-08-26
Add support for the ILI9342C

jessebraham:
First off, thank you for your work on this project! I had previously attempted to implement this driver from scratch and lost motivation, however it was trivial to implement in this repository.

The [ILI9342C](https://focuslcds.com/content/ILI9342C.pdf) is used by the [ESP32-S3-BOX](https://github.com/espressif/esp-box), which is what I used for verification. Based on some very simple testing this seems to be working as intended.

Some notes on this implementation:

- Very similar to the current `ili9486` model, with some small tweaks
- I've based the initialization routine on [this driver](https://github.com/russhughes/ili9342c_mpy); I'm not certain if everything has been covered, but it is funcional
- Currently only supports `Rgb565`; I did try `Rgb666` however the colours were not correct, and I did not spend much time trying to troubleshoot
- For whatever reason the width/height values needed to be the reverse of what I'd normally expect (smaller then larger), however this seems to be how this display works

If you'd like anything changed or added please let me know!

### DIFF
--- a/src/models/ili9342c.rs
+++ b/src/models/ili9342c.rs
@@ -1,5 +1,8 @@
 use display_interface::{DataFormat, DisplayError, WriteOnlyDataCommand};
-use embedded_graphics_core::{pixelcolor::Rgb565, prelude::IntoStorage};
+use embedded_graphics_core::{
+    pixelcolor::{Rgb565, Rgb666},
+    prelude::{IntoStorage, RgbColor},
+};
 use embedded_hal::{blocking::delay::DelayUs, digital::v2::OutputPin};
 
 use crate::{instruction::Instruction, Display, DisplayOptions, Error, Orientation};
@@ -10,6 +13,11 @@ use super::{write_command, Model};
 /// in Rgb565 color mode
 /// Backlight pin is not controlled
 pub struct ILI9342CRgb565;
+
+/// ILI9342C display with Reset pin
+/// in Rgb666 color mode
+/// Backlight pin is not controlled
+pub struct ILI9342CRgb666;
 
 impl Model for ILI9342CRgb565 {
     type ColorFormat = Rgb565;
@@ -37,6 +45,8 @@ impl Model for ILI9342CRgb565 {
 
         delay.delay_us(120_000);
 
+        write_command(di, Instruction::COLMOD, &[0b0101_0101])?; // 16bit 65k colors
+
         init_common(di, delay, options).map_err(|_| Error::DisplayError)
     }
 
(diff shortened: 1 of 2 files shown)


almindor:
Thank you! I'll have a closer look in a day or two.

src/models/ili9342c.rs
```diff
@@ -0,0 +1,109 @@
...
+fn init_common<DELAY, DI>(
+    di: &mut DI,
+    delay: &mut DELAY,
+    options: DisplayOptions,
+) -> Result<u8, DisplayError>
+where
+    DELAY: DelayUs<u32>,
+    DI: WriteOnlyDataCommand,
+{
+    let madctl = options.madctl() ^ 0b0000_1000; // this model has flipped RGB/BGR bit;
```
almindor, line 95: This was a mistake on my end with ST7789. Are you sure it's required for this model?

src/models/ili9342c.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    write_command(di, Instruction::SLPOUT, &[])?; // turn off sleep
+    write_command(di, Instruction::COLMOD, &[0b0101_0101])?; // 16bit 65k colors
```
almindor, line 98: This should be dynamic based on the color type. It's ok for the 565 variant but I wonder if your 666 case didn't work because of this?

I made the same mistake in the ILI9486 driver (I never could've tested rgb565 because they disabled it in hw for the board I have so I just found this bug).

It can probably be made compile time choice tho based on the color type.

almindor (changes requested):
Looks good, just two things I think need attention before we merge it in.

src/models/ili9342c.rs
jessebraham: You're right, this display is in RGB mode by default. For whatever reason my colours still seem to be displaying using BGR (setting the background to red displays as blue) without that, so I think this is why I had set that initially. I will dig into the datasheet a bit more and try to figure out what's going on.

jessebraham: I'm not really sure what's going on here, as far as I can tell the display is configured correctly. I am unable to get it to work properly without inverting this bit, the byte order is always backwards. I did find a couple more drivers for this display, and they both seem to be setting the RGB/BGR bit to 1 (BGR), so I'm not totally crazy at least (well, I hope).

- https://github.com/retro-esp32/Odroid-GO-ILI9342/blob/master/ILI9342/Arduino%20Libraries/Adafruit_ILI9342/Adafruit_ILI9342.cpp#L160  
- https://github.com/MilovdZee/TFTLCD-Library_ILI9342/blob/master/Adafruit_TFTLCD_ILI9342.cpp#L150

Not sure how you'd like to proceed on this. I can update the comment if that is adequate, otherwise I'm not sure what to do.

src/models/ili9342c.rs
jessebraham: Good catch, thank you for pointing that out. Changing to 18bit/262k colours at least made the display not work in a _different_ way using `Rgb666` 😁 I'll keep digging.

src/models/ili9342c.rs
almindor: This is interesting, the first driver inverts RGB but also sets MX bit. The second driver also inverts RGB but then sets the MY bit. So they don't even agree between each other. I guess one is aiming for portrait mirror image, while the other goes for inverted portrait orientation.

almindor (approved):
LGTM, I understand the Rgb666 format still doesn't work right. I'll give it a bit more time so we can look into it but if we get stuck I'll merge as-is.

src/models/ili9342c.rs
almindor: > 

Hmm this would imply some sort of timing issue. What SPI MODE are you using? Does the display have a CS pin?

jessebraham: This ended up being a bit of a red herring.

I was originally testing using `esp32s3-hal` (which is `no_std`), but just as a sanity check I also tested this driver using `esp-idf-hal`/`esp-idf-sys` (which are `std`, a separate project) and this works at all optimization levels. So it appears this issue is somewhere in the `esp-hal` SPI driver.

almindor: Sounds good! It usually is on the SPI level at that point. Thanks again for the model

src/models/ili9342c.rs
```diff
@@ -0,0 +1,193 @@
...
+// simplified constructor for Display
+
+impl<DI, RST> Display<DI, RST, ILI9342CRgb565>
+where
+    DI: WriteOnlyDataCommand,
+    RST: OutputPin,
+{
+    ///
+    /// Creates a new [Display] instance with [ILI9342C] as the [Model]
+    /// *WARNING* Rgb565 only works on non-SPI setups with the ILI9342C!
```
ophilli, line 138: Your pull request description mentions using this on an ESP32-S3-BOX but I also see that you were only able to use 565 and not 666. I thought the box was using SPI - is this documentation that 565 only works with non-SPI setups correct?

rfuest: No, the docs are wrong. The ILI934X controllers support RGB565 and RGB666 via SPI. The comment was probably copied from the ILI9486 support, where it would be correct.

Issue #27, opened 2022-09-04
Support for 240x135 ST7789 displays


andreban:
It seems the library only [supports 240x240 displays](https://github.com/almindor/mipidsi/blob/master/src/models/st7789.rs#L73-L75) at the moment, but there are 240x135 displays available (eg: the [Pimoroni Display Pack](https://shop.pimoroni.com/products/pico-display-pack?variant=32368664215635)).

almindor:
it seems that this is just a cropping effect, see https://github.com/pimoroni/pimoroni-pico/blob/main/drivers/st7789/st7789.cpp#L178-L194

There's a [PR](https://github.com/almindor/mipidsi/pull/20) for a similar display. I think I'll just unify these all into ST7789 model and use a size indicator either in the constructor, or via the DisplayOptions (with sane defaults).

almindor:
@andreban does mipidsi (as is) work with this at all? e.g. does clear() work? Is drawing working ok as long as you use the "cropped" region?

almindor:
@andreban please see if #28 fixes this issue by using the provided `ST7789::st7789_135x240` constructor (or by manually overriding the offset and sizes in `DisplayOptions` if using `with_model`)

andreban:
Thanks for putting together the PR so quickly!

> does mipidsi (as is) work with this at all? e.g. does clear() work? Is drawing working ok as long as you use the "cropped" region?

Yes, and yes.

> please see if https://github.com/almindor/mipidsi/pull/28 fixes this issue by using the provided ST7789::st7789_135x240 constructor (or by manually overriding the offset and sizes in DisplayOptions if using with_model)

Offsets seem a bit off. Here's what's working for me:
```rust
let display_options = DisplayOptions {
    window_offset: (52, 40),
    display_size: (135, 240),
    framebuffer_size: (135, 240),
    ..Default::default()
};
let mut display = Display::st7789_without_rst(lcd_spi_interface, display_options);
```

almindor:
> Thanks for putting together the PR so quickly!
> 
> > does mipidsi (as is) work with this at all? e.g. does clear() work? Is drawing working ok as long as you use the "cropped" region?
> 
> Yes, and yes.
> 
> > please see if #28 fixes this issue by using the provided ST7789::st7789_135x240 constructor (or by manually overriding the offset and sizes in DisplayOptions if using with_model)
> 
> Offsets seem a bit off. Here's what's working for me:
> 
> ```rust
[...]

PR #28, opened 2022-09-05
Refactors the driver and models to support dynamic display sizing and clipping

almindor:
This PR tries to address #27 and #20 by refactoring the `DisplayOptions` and `init` paths so variants of the same `Display` and `Model` can be used with different physical display sizes.

This also adds support for "cropped" display variants in which the display framebuffer is bigger but the visible display is smaller as well as "offset" from the `(0, 0)` coordinates internally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- split [DisplayOptions] into [DisplayOptions] and [ModelOptions] with sizing initialization safety constructors
 - refactored `Display::init` and constructors to match new variant code
 - fixed off by one error in fill operations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- support for model variants via `DisplayOptions`
+- support for `raspberry pico1` variant of the `ST7789` display
+- support for the `waveshare` variants of the `ST7789` display
+
+### Changed
+
+- refactored `Display::init` and constructors to match new variant code
+- fixed off by one error in fill operations
+
+### Removed
+
+- removed "no reset pin" constructor helpers (uses `Option` now)
+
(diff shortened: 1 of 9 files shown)


almindor:
I think I'll make two changes to this PR based on current observations:
1. Change `DisplayOptions::window_offset` into a `Option<Fn(Orientation) -> (u16, u16)>` so that logic can be applied from a closure instead of depending on direct values
2. Change Model constructor logic to enforce "variant constructors" directly so there's no "default" config to fall back to.
3. In relation to 2. probably split `DisplayOptions` so that Model/variant sizing definitions are in `Model` specific options instead.

almindor:
@cloneable @Chippit @andreban @9names this should now be ready to test for all the variants. If you can please check if this code works for you.

It might also be missing some variant constructors, feel free to ping about that. There was also a bug fix in the `fill_solid` function which might explain why lines worked where fills didn't when framebuffer was misdefined (or bugged as it was in this PR before)

9names:
This is what it looks like using st7789_pico1 on commit 442e663
pico_display_new
I'll play with the offsets and see what I come up with...

9names:
Here's the offsets that appear to be correct:
```rust
    match orientation {
        Orientation::Portrait(false) => (52, 40),
        Orientation::Portrait(true) => (53, 40),
        Orientation::Landscape(false) => (40, 52),
        Orientation::Landscape(true) => (40, 53),
        Orientation::PortraitInverted(false) => (53, 40),
        Orientation::PortraitInverted(true) => (52, 40),
        Orientation::LandscapeInverted(false) => (40, 53),
        Orientation::LandscapeInverted(true) => (40, 52),
    }
```

almindor:
@9names @andreban @cloneable @Chippit I think this is really close now. I'm going to wait for a week or so before merging and releasing this in case someone finds an issue since it does change things on the API level

Chippit:
Those Pico1 look good for my Waveshare device as well.

andreban:
I'm using the same Pico display as @9names. These offsets look good to me.

almindor:
Thanks for the testing guys, I'll probably wrap this up in a day or two and release 0.4!

PR #30, opened 2022-10-06
Support for ST7789 with display of 240x240 but frame buffer of 240x320

### LINKED ISSUE #29: Inverted display causes offset in image
It seems part of the screen is not drawn on when using PortraitInverted or LandscapeInverted, as if the image is outside the screen. Though it's fine using Portrait or Landscape. 

I am using `st7789_240x240` method since that is the resolution of the display.

I will contact our provider and try to get the exact part number and datasheet. 
It's supposed to be a st789v2 with 240x240 pixels.

Here are images of the behavior
PXL_20221005_150149231
PXL_20221005_150151669

xgroleau:
Some St7789 have a display of 240x240 but a frame buffer of 240x320 which causes issue when using inverted orientation since
the image is draws in  valid the valid frame buffer pixels, but are not show since they do not exist on the display itself.

Fixes #29. Piggy backed on what was done for the pico display by simply using an offset.

The `st7789_240x240_b240x320` is quite long but I'm not sure how to keep it short and descriptive so I am open to suggestions.

### DIFF
--- a/src/models/st7789/variants.rs
+++ b/src/models/st7789/variants.rs
@@ -66,7 +66,7 @@ where
                 options,
                 (240, 240),
                 (240, 320),
-                st7789_offset,
+                y80_offset,
             )),
         )
     }
@@ -110,8 +110,8 @@ pub(crate) fn pico1_offset(orientation: Orientation) -> (u16, u16) {
     }
 }
 
-// ST7789 pico1 variant with variable offset
-pub(crate) fn st7789_offset(orientation: Orientation) -> (u16, u16) {
+// ST7789 240x240 with a frame buffer of 240x320
+pub(crate) fn y80_offset(orientation: Orientation) -> (u16, u16) {
     match orientation {
         Orientation::Portrait(false) => (0, 0),
         Orientation::Portrait(true) => (0, 0),
--- a/src/models/st7789/variants.rs
+++ b/src/models/st7789/variants.rs
@@ -110,7 +110,8 @@ pub(crate) fn pico1_offset(orientation: Orientation) -> (u16, u16) {
     }
 }
 
-// ST7789 240x240 with a frame buffer of 240x320
+// An offset of 80 y pixels for the st7789 when the display is 240x240 but the frame buffer
+// is 240x320
 pub(crate) fn y80_offset(orientation: Orientation) -> (u16, u16) {
     match orientation {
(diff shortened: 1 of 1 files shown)


src/models/st7789/variants.rs
```diff
@@ -48,6 +48,28 @@ where
...
+    /// Creates a new [Display] instance with [ST7789] as the [Model] with
+    /// general variant using display size of 240x240 but a frame buffer of 240x320 and adjusting the offset
+    ///
+    /// # Arguments
+    ///
+    /// * `di` - a [DisplayInterface](WriteOnlyDataCommand) for talking with the display
+    /// * `rst` - display hard reset [OutputPin]
+    /// * `options` - the [DisplayOptions] for this display/model
+    ///
+    pub fn st7789_240x240_b240x320(di: DI, rst: Option<RST>, options: DisplayOptions) -> Self {
```
almindor, line 61: I'm wondering if the "general" https://github.com/almindor/mipidsi/blob/master/src/models/st7789/variants.rs#L26-L32 ST7789 constructor should be used here and just modified with the offsets.

almindor (commented):
@xgroleau looks good but I think we can use the default st7789 constructor since it would otherwise not serve a role.

Also I'm not 100% sure that this is the right way of doing things. I'll need to dive deeper into the driver specs to see if we're not supposed to use a separate form of "offset" (I feel there's another way to get the same result).

src/models/st7789/variants.rs
xgroleau: The display size woudn't be right if we used the general constructor, though the `st7789_240x240` constructor could be used, but then the frame buffer also wouldn't be right (though maybe it's fine since they shouldn't be used). 

Now as you said maybe it would be possible to use a separate form of offset but I am not familiar with the spec I would need to look it up but it seems the spec is for members only. If there is a cleaner way to do this I am definitely open to it.

almindor: I thought about this a bit more, the issue with another offset setup is that it'd conflict with the clipping offsets.

E.g. I could use the HW scroll to offset the framebuffer "active area" in inverted orientations, and it would work here, but it would break with pico1 or others that have limited display sizes in combination with offset requirements under default conditions.

So I think I will accept this, but I want to test some more first.

almindor (approved):

PR #36, opened 2022-11-03
add invert_colors support

### LINKED ISSUE #33: Inverted Colors with ST7735 Display
I'm using a 128x160 ST7735 display and I'm noticing that the colors appear to be inverted.

` display.clear(Rgb565::WHITE);` makes the whole display black,
`display.set_pixels(1, 1, 10, 10, core::iter::repeat(Rgb565::YELLOW).take(100));` draws a 10x10 red square
`display.set_pixel(11, 11, Rgb565::RED);` draws a single yellow pixel at coordinate (11, 11)

I think that I am setting up the display correctly, below is the code used to create the display as shown in the attached image.

```rust
fn main() -> anyhow::Result<()> {
    // Temporary. Will disappear once ESP-IDF 4.4 is released, but for now it is necessary to call this function once,
    // or else some patches to the runtime implemented by esp-idf-sys might not link properly.
    esp_idf_sys::link_patches();

    let peripherals = Peripherals::take().unwrap();

    let sdo = peripherals.pins.gpio7;
    let sclk = peripherals.pins.gpio6;
    //let cs = peripherals.pins.gpio10;
    let a0 = peripherals.pins.gpio8.into_output()?; // dc
    let rst = peripherals.pins.gpio9.into_output()?;

    let spi = Master::<SPI2, _, _, Gpio2<Unknown>, Gpio10<Unknown>>::new(
        peripherals.spi2,
        Pins {
            sclk,
            sdo,
            sdi: None,
            cs: None,
        },
        //Config::default().baudrate(20_000_000.Hz()),
        Config {
            baudrate: 60_000_000.Hz(),
            write_only: true,
            ..Default::default()
        },
    )?;

    let di = display_interface_spi::SPIInterfaceNoCS::new(spi, a0);
    let mut display = mipidsi::Builder::st7735s(di)
        .with_display_size(128, 160)
        .with_framebuffer_size(128, 160)
        .with_color_order(mipidsi::ColorOrder::Rgb)
        .with_orientation(mipidsi::Orientation::Portrait(false))
        .init(&mut Ets, Some(rst))
        .unwrap();

    display.clear(Rgb565::WHITE);

    display.set_pixels(1, 1, 10, 10, core::iter::repeat(Rgb565::YELLOW).take(100));
    display.set_pixel(11, 11, Rgb565::RED);

    Ok(())
}
```
IMG_20221101_192624

almindor:
Adds `Builder::invert_colors(bool)` to fix #33 and surplant #34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+ - added `ModelOptions::invert_colors` flag
+ - added `Builder::with_invert_colors(bool)` method
+
+### Changed
+
+ - `Model::init` changed to expect `options: &ModelOptions`
+
 ### Removed
 
  - removed duplicated `INVON` call in `ST7735s` model init
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,15 @@
 # Migration guide for MIPIDSI driver
 
+## v0.5 -> 0.6
+
+### Users
+
+* no change, addition of `Builder::with_invert_colors(bool)` allows more combinations of display variants.
+
+### Model writers and specific variants
+
+`Model::init` now expects the `options: &ModelOptions` instead of just `madcl: u8` argument. This allows the use of the `invert_colors` field during init.
+
 ## v0.4 -> 0.5
 
 ### Users
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -45,6 +45,14 @@ where
     }
 
     ///
+    /// Sets the invert color flag
+    ///
+    pub fn with_invert_colors(mut self, invert_colors: bool) -> Self {
+        self.options.invert_colors = invert_colors;
+        self
+    }
+
+    ///
     /// Sets the [ColorOrder]
     ///
     pub fn with_color_order(mut self, color_order: ColorOrder) -> Self {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,7 +84,7 @@ where
         delay_source: &mut impl DelayUs<u32>,
     ) -> Result<u8, InitError<RST::Error>> {
         self.model
-            .init(&mut self.di, delay_source, self.madctl, &mut self.rst)
+            .init(&mut self.di, delay_source, &self.options, &mut self.rst)
     }
     ///
     /// Returns currently set [Orientation]
--- a/src/models.rs
+++ b/src/models.rs
@@ -23,7 +23,7 @@ pub trait Model {
         &mut self,
         di: &mut DI,
         delay: &mut DELAY,
-        madctl: u8,
+        options: &ModelOptions,
         rst: &mut Option<RST>,
     ) -> Result<u8, InitError<RST::Error>>
     where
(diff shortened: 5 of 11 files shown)


almindor:
@cdsupina could you please try this PR using `Builder::with_color_order(ColorOrder::bgr)` and `Builder::with_invert_colors(true)` to see if it works ok ?

Thanks

cdsupina:
I tried this branch and it appears like using just `Builder::with_color_order(ColorOrder::Bgr)` makes the colors look as they should. If I also use `Builder::with_invert_colors(true)`, the colors are inverted.

almindor:
> I tried this branch and it appears like using just `Builder::with_color_order(ColorOrder::Bgr)` makes the colors look as they should. If I also use `Builder::with_invert_colors(true)`, the colors are inverted.

Hmm I wonder if issuing `INVOFF` is what does it in this case. I made it so that either `INVOFF` or `INVON` now gets called depending on what was configued (default would be `INVOFF`). It seems your display simply has `INVON` by default. Could you try commenting out line 39 of `src/models/st7735s.rs` so that neither is called and see what happens?

rfuest:
> Hmm I wonder if issuing `INVOFF` is what does it in this case. I made it so that either `INVOFF` or `INVON` now gets called depending on what was configued (default would be `INVOFF`). It seems your display simply has `INVON` by default. Could you try commenting out line 39 of `src/models/st7735s.rs` so that neither is called and see what happens?

I don't think this is the case and I expect that `INVOFF` and no `INV..` command will behave the same. The difference between inverted and non inverted display seems to be related to the LCD glass itself. I believe that TN and IPS panels have different default states, where one lets light through if no voltage is applied and the other one blocks the light with no voltage.

almindor:
> > Hmm I wonder if issuing `INVOFF` is what does it in this case. I made it so that either `INVOFF` or `INVON` now gets called depending on what was configued (default would be `INVOFF`). It seems your display simply has `INVON` by default. Could you try commenting out line 39 of `src/models/st7735s.rs` so that neither is called and see what happens?
> 
> I don't think this is the case and I expect that `INVOFF` and no `INV..` command will behave the same. The difference between inverted and non inverted display seems to be related to the LCD glass itself. I believe that TN and IPS panels have different default states, where one lets light through if no voltage is applied and the other one blocks the light with no voltage.

No, you're right. I got confused thinking @cdsupina 's test was with INVOFF + BGR but he said INVON + BGR was wrong, which is correct. So this should now be solved. I will do more testing with my own displays to make sure it works right there as well and merge after.

cdsupina:
Thanks for tackling this issue. Let me know if you need any more tests from me.

almindor:
> Thanks for tackling this issue. Let me know if you need any more tests from me.

Sure thing. I have "a" solution ready, but I think I'll refactor things a bit more. Feel free to use the PR for now so you don't get stuck on your end. The final version should be compatible from the user's perspective.

almindor:
@cdsupina this should now no longer require passing the invert flag for ST7789 and ST7735s display models as that seems to be the default. Still a WIP as I'm not happy with the init code atm. will want to try and refactor things into something more readable.

almindor:
I'm going to merge this so we can concentrate on refactoring without having multiple PRs around. This won't get released just yet though. @cdsupina I hope this is ok for now. I'm hoping to have the final next release done in a week or two at worst, you can use this PR or the master branch in the meantime. Things shouldn't change much API wide (we might introduce the new orientation changes into this release though)

PR #39, opened 2022-11-06
Add ILI9341 and use display-driver-hal

rfuest:
The PR adds ILI9341 support. I've marked it as a draft PR because it also uses the unreleased `display-driver-hal` to test it's API. The offset calculation is also disabled at the moement, because I didn't have an appropriate display to test it.

The new `Orientation` type uses a builder like interface:
```rust
let display = Builder::ili9341_rgb565(interface)
    .with_orientation(Orientation::new().rotate(Rotation::Deg270).flip_horizontal())
    .init(&mut delay, Some(display_reset))
    .unwrap();
```
IMO this is easier to use, because you can easily tweak the orientation by adding a `rotate` or `flip_vertical`/`flip_horizontal` commands. Perviously it was only possible to mirror around one axis and the user had to add the necessary rotation manually if they wanted to mirror around the other axis.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ display-interface = "0.4.1"
 embedded-graphics-core = "0.3.3"
 embedded-hal = "0.2.7"
 nb = "1.0.0"
+display-driver-hal = { git = "https://github.com/embedded-graphics/display-driver-hal.git" }
 
 [dependencies.heapless]
 optional = true
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,8 +29,12 @@ pub mod instruction;
 
 use crate::instruction::Instruction;
 
+use display_driver_hal::MemoryMapping;
+use display_driver_hal::Orientation;
 use display_interface::DataFormat;
 use display_interface::WriteOnlyDataCommand;
+
+pub use display_driver_hal as hal;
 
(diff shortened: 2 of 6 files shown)


georgik:
Driver works on ESP32, tested with M5 Core Fire and Wokwi. Color channels are also correct for RGB565. 👍

almindor:
Looks good. NOTE: I'm also in the middle of https://github.com/almindor/mipidsi/pull/44 but I think the orientation change would be easier to get in first.

rfuest:
I can finish the PR if you want to merge it before #44, but this would also mean that `display-driver-hal` needs to be released on crates.io before the next release of `mipidsi`.

almindor:
> I can finish the PR if you want to merge it before #44, but this would also mean that `display-driver-hal` needs to be released on crates.io before the next release of `mipidsi`.

What if we split this into 2 PRs then? get the model in ASAP, and then we can do the orientation separately.

rfuest:
> What if we split this into 2 PRs then? get the model in ASAP, and then we can do the orientation separately.

OK, I'll prepare another PR without the orientation changes.

rfuest:
The ILI9341 model has already been added in #45 and the rest of the changes are now included in #80.

PR #40, opened 2022-11-08
Update lib.rs

lemonxah:
the batch code is not reachable


lemonxah:
already calls draw batch in the draw iter if the feature is enabled, i wanted to try use draw batch cause the draw iter isn't rendering to the screen. let me try without the batch feature and see if that is working

PR #41, opened 2022-11-19
Update embedded-hal to 1.0.0-alpha.8

mixedCase:


almindor:
This looks good! I'll keep it as a PR for a bit until my refactor is done. We'll have to revisit once that lands and then a pre-release could be made for simpler usage.

almindor:
This has been superseded

PR #45, opened 2022-12-12
Add ILI9341 model

rfuest:
This PR adds the ILI9341 model, that was previously added as part of #39. I've cleaned up the code some more and extracted the common code for the `ILI3941` and `ILI9342C` controllers into a separate module.

~~One remaining issue is the color order: Before #36 the RGB/BGR bit for the ILI9342C was inverted. According to the datasheet this shouldn't be necessary, but without the inversion I get incorrect colors on my display. I'm not sure if the datasheet is wrong or what is going on, but I've checked that my display is indeed using the RGB subpixel order by using a magnifying glass.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,16 +9,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
- - added `ModelOptions::invert_colors` flag
- - added `Builder::with_invert_colors(bool)` method
+- added `ModelOptions::invert_colors` flag
+- added `Builder::with_invert_colors(bool)` method
+- added `ILI9341` model support
 
 ### Changed
 
- - `Model::init` changed to expect `options: &ModelOptions`
+- `Model::init` changed to expect `options: &ModelOptions`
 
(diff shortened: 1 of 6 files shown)


rfuest:
Forget what I wrote about the color order, my test program was broken.

rfuest:
@georgik Could you test if this PR works as well as #39 for your display? Or if you need to set the color order to BGR to get it working?

georgik:
@rfuest Thank you for the update. 

I've run the tests and it seems to suffer the same problem like my PR when I copied the code from `ILI9342C`. Blue and red chnnel seems to be swapped. The version on PR #39 works.

Incorrect version
image


Correct version

georgik:
@rfuest Thank you very much for the suggestion. Using explicit `with_color_order` seems to be good option, because there are different displays with different color channel order. E.g. Kaluga is using RGB while M5Stack is using BGR.

rfuest:
@georgik Good to hear that it works for you.
@almindor The PR is ready to review/merge now.

almindor:
Thank you

PR #50, opened 2023-01-12
add `Builder::with_window_offset_handler` method

lily-mara:
`ModelOptions` has a field for storing an offset handler, but it's currently not accessible unless you provide a custom `Model` implementation to override `Model::default_options`. This addition allows you to set the offset handler via the `Builder`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- added `Builder::with_window_offset_handler` method
 - added `ModelOptions::invert_colors` flag
 - added `Builder::with_invert_colors(bool)` method
 - added `ILI9341` model support
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -101,6 +101,17 @@ where
     }
 
     ///
+    /// Sets the window offset handler
+    ///
+    pub fn with_window_offset_handler(
+        mut self,
+        window_offset_handler: fn(_: &ModelOptions) -> (u16, u16),
+    ) -> Self {
+        self.options.window_offset_handler = window_offset_handler;
+        self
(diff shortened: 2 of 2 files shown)


almindor (approved):
Thank you!

lily-mara:
Thanks for the quick review! Do you have plans to cut a release any time soon? The display I'm working with requires an offset handler and it would be nice to remove my custom `Model` impl

almindor:
> Thanks for the quick review! Do you have plans to cut a release any time soon? The display I'm working with requires an offset handler and it would be nice to remove my custom `Model` impl

Hmm so the main blocker right now is #44 which needs more work. Would a pre-release work for you?

lily-mara:
Yes, a pre-release would be just fine for me! What is the nature of the DCS changes, is the crate usable with the state of that PR right now?

almindor:
> Yes, a pre-release would be just fine for me! What is the nature of the DCS changes, is the crate usable with the state of that PR right now?

The DCS changes rewrites all the commands into more abstracted (and safer) versions. This will be offloaded to a separate "mipidcs" crate at some point. It changes all the inits, but in a compatible way. Basically instead of say "write_command(di, Instruction::MADCTL, &[some_data])` you can now do things like `dcs.set_address_mode(address_mode_command)` where `address_mode_command` consists all the options for `MADCTL` in a safe abstracted way. (similar to how memory mapped registers are abstracted in a PAC/HAL).

It's also the first step in adding read functionality to displays although not really a blocker for it.

almindor:
Looking at the current unreleased list in the changelog I think I'll do a normal 0.6 (since API changed) release for what we have in master right now. The DCS stuff can be 0.7

almindor:
v0.6 is out https://crates.io/crates/mipidsi

PR #52, opened 2023-02-19
Documentation improvements

rfuest:
While I was looking at the new code I discovered that some parts of the code weren't documented yet. This PR does at least add some basic documentation to all parts of the code.

I've changed the style of documentation a bit to use the style recommended in the Rust style guide and API guidelines. While the preferred commenting style is subjective there are some objective advantages of using the recommended style. Especially using a summary line (https://doc.rust-lang.org/1.0.0/style/style/comments.html#summary-line) makes the generated docs more readable.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -116,7 +116,8 @@ where
 }
 
 impl BitsPerPixel {
-    pub const fn from_rgbcolor<C: RgbColor>() -> Self {
+    /// Returns the bits per pixel for a embedded-graphics [`RgbColor`].
+    pub const fn from_rgb_color<C: RgbColor>() -> Self {
         let bpp = C::MAX_R.trailing_ones() + C::MAX_G.trailing_ones() + C::MAX_B.trailing_ones();
 
         match bpp {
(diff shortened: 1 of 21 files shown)


src/graphics.rs
```diff
@@ -116,7 +116,8 @@ where
 }
 
 impl BitsPerPixel {
-    pub const fn from_rgbcolor<C: RgbColor>() -> Self {
+    /// Returns the bits per pixel for a embedded-graphics [`RgbColor`].
+    pub const fn from_rgb_color<C: RgbColor>() -> Self {
```
rfuest, line 120: Not related to the docs, but at least IMO more readable and the default conversion from camel case to snake case.

src/options.rs
```diff
@@ -60,24 +55,22 @@ impl ModelOptions {
         }
     }
 
-    pub fn with_invert_colors(mut self, color_inversion: ColorInversion) -> Self {
+    /// Sets the color inversion setting.
+    pub fn set_invert_colors(&mut self, color_inversion: ColorInversion) {
```
rfuest, line 59: I did change this to match `set_orientation`. Having two different methods of settings fields in one type seemed confusing.

almindor:
This is great! thank you

PR #54, opened 2023-03-01
A Add pixecolor Rgb888 for TGA file upload

FMFigueroa:


almindor (changes requested):

almindor:
I think this looks good apart from the formatting and `write_pixels_rgb888` being wrong. AFAICS it should just send the color bytes in directly.

FMFigueroa:
with reason the image is overwritten, buffer is collapsed, I didn't know this, thanks for explaining it @rfuest , I was looking to transfer an image file with the tinytga crate doing a bypass with the driver

rfuest:
> with reason the image is overwritten, buffer is collapsed, I didn't know this, thanks for explaining it @rfuest , I was looking to transfer an image file with the tinytga crate doing a bypass with the driver.
> I will do it with a bmp image that requires Rgb565 and is supported by ili9341.

Using an image with the same color format will always give the best performance, but you can still use `tinytga` if you like, for example for the RLE compression.

There are two ways to use use `tinytga` in your case: Either use `color_converted` (see https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTargetExt.html#tymethod.color_converted), or preferably the GIT version of `tinytga`, which automatically converts mismatched color formats. Not quite sure why the latest changes in `tinytga` weren't released yet, like they were for `tinybmp`.

Issue #56, opened 2023-05-09
ST7789 display is black with mipidsi crate, until program execution is interrupted- working fine with st7789 crate


tingox:
This example
```
// nrf52840dk-mipidsi_examples - xtext
//
#![no_std]
#![no_main]

extern crate cortex_m_rt as rt;

extern crate nrf52840_hal;
extern crate panic_halt;

use cortex_m_rt::entry;
use defmt_rtt as _; // global logger
use display_interface_spi::SPIInterfaceNoCS;
use embedded_graphics::{
    mono_font::{ascii::{FONT_6X10, FONT_10X20}, MonoTextStyle},
    pixelcolor::Rgb565,
    prelude::*,
    text::Text,
};

use nrf52840_hal::gpio::p0::Parts;
use nrf52840_hal::gpio::Level;
use nrf52840_hal::prelude::*;
use nrf52840_hal::spim;
use nrf52840_hal::Delay;

use mipidsi::{Builder, Orientation};

#[entry]
fn main() -> ! {
    let core = nrf52840_hal::pac::CorePeripherals::take().unwrap();
    let mut delay = Delay::new(core.SYST);

    let p = nrf52840_hal::pac::Peripherals::take().unwrap();
    let port0 = Parts::new(p.P0);

    let _backlight = port0.p0_31.into_push_pull_output(Level::Low); // set medium backlight on
    // let rst = port0.p0_29.into_push_pull_output(Level::Low); // reset pin
    let rst = port0.p0_29.into_push_pull_output(Level::High);   // reset pin
    let _cs = port0.p0_28.into_push_pull_output(Level::Low);    // keep low while driving display
    let dc = port0.p0_30.into_push_pull_output(Level::Low);     // data/clock switch

    let spiclk = port0.p0_03.into_push_pull_output(Level::Low).degrade(); // SPI clock to LCD
    let spimosi = port0.p0_04.into_push_pull_output(Level::Low).degrade(); // SPI MOSI to LCD

    let pins = spim::Pins {
        sck: spiclk,
        miso: None,
        mosi: Some(spimosi),
    };

    // create SPI interface
    let spi = spim::Spim::new(p.SPIM0, pins, spim::Frequency::M8, spim::MODE_3, 122);

    // display interface abstraction from SPI and DC
    let di = SPIInterfaceNoCS::new(spi, dc);

    // create driver
    let mut display = Builder::st7789(di)          // known model
            .with_display_size(240, 240)           // set options
[...]

PR #59, opened 2023-05-18
add better info about backlight and MIPI DCS/DSI

almindor:
Clarifies naming of the crate as well as backlight manual control requirement.

See #56 for reasons behind clarifications.


almindor:
@tingox @rfuest do you think this works or should I add more info in the docs somewhere as well?

tingox:
Two things tripped me up:
1) the display I was using doesn't show anything unless the backlight is on
2) the older crate (st7789) automatically controlled the backlight pin
neither is something that can be blamed on this crate.
But perhaps a troubleshooting hint might be helpful?
Something like: "if you can't see anything on your display, try turning on / activating the backlight"

README.md
```diff
@@ -1,12 +1,18 @@
 # mipidsi
 
 This crate provides a generic display driver to connect to TFT displays
-that implement the [MIPI DSI](https://www.mipi.org/specifications/dsi).
+that implement the [MIPI DCS](https://www.mipi.org/specifications/display-command-set).
```
rfuest, line 4:
```suggestion
that implement the [MIPI Display Command Set (DCS)](https://www.mipi.org/specifications/display-command-set).
```

rfuest:
I would move the backlight note into a new "Troubleshooting" section, which would make it stand out more than being part of the notes list.

Another useful addition to the README would be a link to the API docs, which I've missed before. Perhaps in the form of some badges to docs.rs and crates.io.

almindor:
> I would move the backlight note into a new "Troubleshooting" section, which would make it stand out more than being part of the notes list.
> 
> Another useful addition to the README would be a link to the API docs, which I've missed before. Perhaps in the form of some badges to docs.rs and crates.io.

Good calls, added badges and troubleshooting section.

README.md
```diff
@@ -1,12 +1,22 @@
 # mipidsi
 
+<p align="center">
+    <a href="https://crates.io/crates/mipidsi"><img src="https://img.shields.io/crates/v/mipidsi.svg" alt="Crates.io"></a>
+    <a href="https://docs.rs/mipidsi"><img src="https://docs.rs/mipidsi/badge.svg" alt="Docs.rs"></a>
+    <a href="https://matrix.to/#/#rust-embedded-graphics:matrix.org"><img src="https://img.shields.io/matrix/rust-embedded-graphics:matrix.org" alt="Discuss driver on embedded-graphics on Matrix"></a>
+</p>V
```
rfuest, line 7:
```suggestion
</p>
```

rfuest:
Do you also want to update `lib.rs` in this PR or open another one? IMO the docs should also contain a link to the troubleshooting doc.

almindor:
> Do you also want to update `lib.rs` in this PR or open another one? IMO the docs should also contain a link to the troubleshooting doc.

should be done now, the troubleshooting link from docs.rs will only work once merged of course

PR #60, opened 2023-05-28
Async

monacoprinsen:
Update to work with async functionality and the embedded hal alpha 10


almindor:
I've split the repo into a workspace with mipidsi-async being a separate crate. Please put any async work there.

almindor:
@monacoprinsen any plans on this?

monacoprinsen:
This branch works great in async mode!

techmccat:
> I've split the repo into a workspace with mipidsi-async being a separate crate. Please put any async work there.

Sorry to bump this much later, wouldn't it be better to use something like maybe-async to avoid all the duplication? It's mostly just going to be converting sync bus writes to async ones as far as i understand.

almindor:
> > I've split the repo into a workspace with mipidsi-async being a separate crate. Please put any async work there.
> 
> Sorry to bump this much later, wouldn't it be better to use something like maybe-async to avoid all the duplication? It's mostly just going to be converting sync bus writes to async ones as far as i understand.

The biggest issue is that embedded-graphics still isn't  async. This means that even if we add async support, only users that would do `async set_pixel` calls would be actually using it. Nobody that interfaces via e-g would go that route.

I don't think it's worth the effort until the e-g ecosystem becomes async.

techmccat:
Having worked a bit with the small page-mode monochrome displays I'd say that even just async framebuffer flushing could be rather nice to have, as you've mentioned in #171 .

Issue #48, opened 2022-12-15
Remove color inversion form `default_options`


rfuest:
The `default_options` for some models contain `with_invert_colors(true)` and others don't. IMO it would be better to be consistent across all models and default to non inverted colors.

The decision if color inversion is necessary is entirely dependent on the LCD glass and not the controller model and can only be decided for a specific LCD module. For LCD module specific constructors, like the `st7789_pico1`, the `with_invert_colors(true)` should stay. To help users who want to use other LCD modules there should be additional docs to help them identify problems, as also mentioned in #47.

almindor:
Good call, I'll make sure to include in the next api breaking release.

rfuest:
The color inversion has been removed from the default options in #80.

PR #61, opened 2023-06-09
Add GC9A01 model

andelf:
Tested with a round 240x240 display with GC9A01 controller.

Testing code:

```rust
    let mut display = mipidsi::Builder::gc9a01(di)
        .with_display_size(240, 240)
        .with_color_order(mipidsi::ColorOrder::Bgr)
        .with_invert_colors(mipidsi::ColorInversion::Inverted)
        .init(&mut delay, Some(rst))
        .unwrap();
    display.set_orientation(mipidsi::Orientation::Portrait(false));
    display.clear(Rgb565::WHITE).unwrap();
```

### DIFF
--- a/src/models/gc9a01.rs
+++ b/src/models/gc9a01.rs
@@ -1,0 +1,167 @@
+use display_interface::{DataFormat, WriteOnlyDataCommand};
+use embedded_graphics_core::{pixelcolor::Rgb565, prelude::IntoStorage};
+use embedded_hal::{blocking::delay::DelayUs, digital::v2::OutputPin};
+
+use crate::{
+    dcs::{
+        BitsPerPixel, ExitSleepMode, PixelFormat, SetAddressMode, SetDisplayOn, SetInvertMode,
+        SetPixelFormat, SoftReset, WriteMemoryStart,
+    },
+    error::InitError,
+    Builder, ColorInversion, Error, ModelOptions,
+};
+
+use super::{Dcs, Model};
+
+/// GC9A01 display in Rgb565 color mode.
+pub struct GC9A01;
+
+impl Model for GC9A01 {
+    type ColorFormat = Rgb565;
+
+    fn init<RST, DELAY, DI>(
+        &mut self,
+        dcs: &mut Dcs<DI>,
+        delay: &mut DELAY,
+        options: &ModelOptions,
+        rst: &mut Option<RST>,
+    ) -> Result<SetAddressMode, InitError<RST::Error>>
+    where
+        RST: OutputPin,
+        DELAY: DelayUs<u32>,
+        DI: WriteOnlyDataCommand,
+    {
+        let madctl = SetAddressMode::from(options);
+
+        match rst {
+            Some(ref mut rst) => self.hard_reset(rst, delay)?,
+            None => dcs.write_command(SoftReset)?,
+        }
+        delay.delay_us(200_000);
+
+        dcs.write_raw(0xEF, &[])?; // inter register enable 2
+        dcs.write_raw(0xEB, &[0x14])?;
+        dcs.write_raw(0xFE, &[])?; // inter register enable 1
+        dcs.write_raw(0xEF, &[])?; // inter register enable 2
+        dcs.write_raw(0xEB, &[0x14])?;
+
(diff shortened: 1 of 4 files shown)


src/models/gc9a01.rs
```diff
@@ -0,0 +1,171 @@
...
+
+use super::{Dcs, Model};
+
+/// GC9A01 display in Rgb565 color mode.
+pub struct GC9A01;
+
+impl Model for GC9A01 {
+    type ColorFormat = Rgb565;
+
+    fn init<RST, DELAY, DI>(
```
rfuest, line 22: Did you copy the init sequence from example code from a display manufacturer or somewhere else? I'm curious because the code contains a lot of commands which aren't mentioned in the GC9A01A datasheet I found. This could just be a difference between GC9A01 and GC9A01A or the code contains some invalid commands.

src/models/gc9a01.rs
```diff
@@ -0,0 +1,171 @@
...
+            &[0x3C, 0x00, 0xCD, 0x67, 0x45, 0x45, 0x10, 0x00, 0x00, 0x00],
+        )?;
+        dcs.write_raw(
+            0x67,
+            &[0x00, 0x3C, 0x00, 0x00, 0x00, 0x01, 0x54, 0x10, 0x32, 0x98],
+        )?;
+
+        dcs.write_raw(0x74, &[0x10, 0x85, 0x80, 0x00, 0x00, 0x4E, 0x00])?;
+        dcs.write_raw(0x98, &[0x3e, 0x07])?;
+        dcs.write_raw(0x35, &[0x00])?; // tearing effect line ON
```
rfuest, line 121: The tearing effect output shouldn't be enabled in the init sequence. It can be controlled by: https://docs.rs/mipidsi/latest/mipidsi/struct.Display.html#method.set_tearing_effect

```suggestion
```

almindor (changes requested):
As @rfuest said, we want to keep the tearing effect as a user choice. Otherwise looks good for a first model implementation. I would like to de-raw-ify the DCS commands and have those that are missing be implemented, but that can be done as a separate step.

Also please add a CHANGELOG entry and fix formatting with `cargo fmt`

src/models/gc9a01.rs
```diff
@@ -0,0 +1,171 @@
...
+
+// simplified constructor on Display
+
+impl<DI> Builder<DI, GC9A01>
+where
+    DI: WriteOnlyDataCommand,
+{
+    /// Creates a new display builder for GC9A01 displays in Rgb565 color mode.
+    ///
+    /// The default framebuffer size is 132x162 pixels and display size is 80x160 pixels.
```
rfuest, line 162: This comment needs to be updated.

src/models/gc9a01.rs
```diff
@@ -0,0 +1,171 @@
...
+        let mut iter = colors.into_iter().map(|c| c.into_storage());
+
+        let buf = DataFormat::U16BEIter(&mut iter);
+        dcs.di.send_data(buf)?;
+        Ok(())
+    }
+
+    fn default_options() -> ModelOptions {
+        let mut options = ModelOptions::with_sizes((240, 240), (240, 240));
+        options.set_invert_colors(ColorInversion::Inverted);
```
rfuest, line 148: @almindor Should we remove this now or later? We had planned to remove the default inversions in issue #48.

rfuest (changes requested):

src/models/gc9a01.rs
almindor: Yah I think new models should be set without inversions.

src/models/gc9a01.rs
andelf: They are from manufacturer-provided codes.
There are some variants with minor differences.
See-also https://github.com/adafruit/Adafruit_GC9A01A/blob/main/Adafruit_GC9A01A.cpp

src/models/gc9a01.rs
andelf: done

almindor (changes requested):
Just needs a CHANGELOG entry now

almindor (approved):
LGTM

Issue #64, opened 2023-07-07
ILI9341 not working with RPI Pico


Slushee-a:
I wanted to drive the ILI9341 display with a Raspberry Pi Pico using the mipidsi crate, but it doesn't work. The screen stays white. I've searched the internet to try to find what my problem is but I couldn't find anything. The documentation is also very limited and there are no code examples (the one in the docs isn't even correct, `.init()` returns a result) and no wiring examples. 

It's probably not working because I might have missed something from the docs or wired something wrong. But I can't for the life of me figure out where the problem is.

I used [**this carrier board »**](https://www.az-delivery.de/en/products/2-4-tft-lcd-touch-display) with [**this wiring »**](https://github.com/almindor/mipidsi/assets/55996847/ff5b1c8e-d518-4f26-ab9f-5efa89a9dd5c).

Here is a table with the connections, in case my drawing of the wiring isn't clear enough:
| Board   | Standard | Reasoning   | Connected to | Function      |
| ------- | -------- | ----------- | ------------ | ------------- |
| LCD_RD  | SPI MOSI | RD = Read   | GP15         | SPI TX (MOSI) |
| LCD_WR  | SPI MISO | WR = Write  | GP12         | SPI RX (MISO) |
| LCD_RS  | D/C      | Guess       | GP14         | Digital out   |
| LCD_CS  | CS       | CS = Chip Select | GP13 | Digital out   |
| LCD_RST | RESET    | RST = Reset | GP11         | Digital out   |
| 5V      |          |             | VSYS         |               |
| 3V3     |          |             | 3V3 (OUT)    |               |
| GND     |          |             | GND          |               |

This is the code I used: [**Snippet »**](https://gitlab.com/-/snippets/2567289)

If on the loop I add the following:
```rust
loop {
    display.clear(Rgb666::BLUE).unwrap();
    delay.delay_ms(100);
}
```

The oscilloscope reads the following:
| Red   | Blue |
| ----- | ---- |
| Clock | MOSI |

![image](https://github.com/almindor/mipidsi/assets/55996847/8b282a62-d1c1-4aeb-810f-397a79c4e02d)

| Red   | Blue |
| ----- | ---- |
| Clock | MISO |

![image](https://github.com/almindor/mipidsi/assets/55996847/fb07e2cb-08e0-460d-b3ce-e14c4656c8e0)

The MOSI pin doesn't seem to be sending any important data, and the signal on the MISO pin looks really weird.

**EDIT:**
[...]

PR #65, opened 2023-07-07
Amplify documentation and add example

Slushee-a:
Made this merge request as mentioned in #64.

# Why
Coming into the crate I wasn't sure where `DisplayInterface` and `Rgb565` came from, as well as which modules I needed to import. Browsing through different crate docs and looking at clippy helps you figure out all of that, but adding more, and less simplified examples saves you most of that trouble.

# Changes included
 - Amplify (and fix) existing example on docs homepage
 - Add example on docs homepage
 - Add list of `display-interface` crates that can be used to create the `DisplayInterface`
 - Add usage example to `.orientation()` and `.setpixel()`
 - Add usage example to `Builder` docs page
 - Add an `/examples` folder (Contains an example for the pi pico with the ili9341)
 
# Could be improved:
 - Missing usage example for [`display.set_pixels()`](https://docs.rs/mipidsi/latest/mipidsi/struct.Display.html#method.set_pixels) (It's not very clear what "*anything that can provide `IntoIterator<Item = u16>`*" is supposed to be)
 - Explanation on [`display.set_scroll_region()`](https://docs.rs/mipidsi/latest/mipidsi/struct.Display.html#method.set_scroll_region) isn't very extensive/clear
 - Another example on the `/examples` folder with another display, microcontroller and interface used. Could also include a more advanced drawing rather than just the whole screen being one color. A colored square would be nice.

### DIFF
--- a/examples/parallel_ili9341_rp_pico/.cargo/config.toml
+++ b/examples/parallel_ili9341_rp_pico/.cargo/config.toml
@@ -1,0 +1,32 @@
+[target.'cfg(all(target_arch = "arm", target_os = "none"))']
+# Choose a default "cargo run" tool:
+# - probe-run provides flashing and defmt via a hardware debugger
+# - cargo embed offers flashing, rtt, defmt and a gdb server via a hardware debugger
+#     it is configured via the Embed.toml in the root of this project
+# - elf2uf2-rs loads firmware over USB when the rp2040 is in boot mode
+# runner = "probe-run --chip RP2040"
+# runner = "cargo embed"
+runner = "elf2uf2-rs -d"
+
+rustflags = [
+  "-C", "linker=flip-link",
+  "-C", "link-arg=--nmagic",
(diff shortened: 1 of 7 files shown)


Slushee-a:
I see what went wrong. 
1. Some code on the codeblock goes over the standard formatting limit
2. I didn't ignore the codeblocks

I will fix this in my next commit very soon

Slushee-a:
I'm really sorry... Apparently it failed because some trailing spaces? otherwise I can't see what made the pipeline failed. Fixed in the latest commit.

almindor:
> I'm really sorry... This is embarassing. Sorry for wasting your time. Apparently it failed because some trailing spaces? otherwise I can't see what made the pipeline failed. Fixed in the latest commit.

No problem at all. You're giving something to improve the library, there's nothing to feel sorry or embarrased about.

almindor (approved):
Thanks for all the improvements!

PR #66, opened 2023-07-08
Add example on docs and /examples folder

Slushee-a:
# Why
As mentioned in #65, two improvements that could be made were adding another example project and expanding documentation on [`set_pixels()`](https://docs.rs/mipidsi/latest/mipidsi/struct.Display.html#method.set_pixels)

# Changes included
 - Adds another example to the `/examples` folder. 
   - Made for the esp32-c3
   - Using the ili9486 display with SPI
   - Shows a basic example of how[`set_pixels()`](https://docs.rs/mipidsi/latest/mipidsi/struct.Display.html#method.set_pixels) can be used
   
 - Adds an example to the [`set_pixels()`](https://docs.rs/mipidsi/latest/mipidsi/struct.Display.html#method.set_pixels) on the docs

### DIFF
--- a/examples/spi-ili9486-esp32-c3/src/main.rs
+++ b/examples/spi-ili9486-esp32-c3/src/main.rs
@@ -1,0 +1,364 @@
+#![no_std]
+#![no_main]
+
+/* --- Needed by ESP32-c3 --- */
+use esp_backtrace as _;
+use hal::{
+    clock::ClockControl,
+    peripherals::Peripherals,
+    prelude::*,
+    spi::{Spi, SpiMode},
+    timer::TimerGroup,
+    Delay, Rtc, IO,
+};
+/* -------------------------- */
+
+use embedded_graphics::{
+    // Provides the necessary functions to draw on the display
+    draw_target::DrawTarget,
+    // Provides colors from the Rgb565 color space
+    pixelcolor::Rgb565,
+    prelude::RgbColor,
+};
+
+// Provides the parallel port and display interface builders
+use display_interface_spi::SPIInterfaceNoCS;
+
+// Provides the Display builder
+use mipidsi::Builder;
+
+use fugit::RateExtU32;
+
+#[entry]
+fn main() -> ! {
+    let peripherals = Peripherals::take();
+    let mut system = peripherals.SYSTEM.split();
+    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
+    let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
+
+    // Disable the RTC and TIMG watchdog timers
+    let mut rtc = Rtc::new(peripherals.RTC_CNTL);
+    let timer_group0 = TimerGroup::new(
+        peripherals.TIMG0,
+        &clocks,
+        &mut system.peripheral_clock_control,
+    );
+    let mut wdt0 = timer_group0.wdt;
+    let timer_group1 = TimerGroup::new(
+        peripherals.TIMG1,
+        &clocks,
+        &mut system.peripheral_clock_control,
+    );
+    let mut wdt1 = timer_group1.wdt;
+    rtc.swd.disable();
+    rtc.rwdt.disable();
+    wdt0.disable();
+    wdt1.disable();
+
+    // Define the delay struct, needed for the display driver
+    let mut delay = Delay::new(&clocks);
+
+    // Define the Data/Command select pin as a digital output
+    let dc = io.pins.gpio7.into_push_pull_output();
+    // Define the reset pin as digital outputs and make it high
+    let mut rst = io.pins.gpio8.into_push_pull_output();
+    rst.set_high().unwrap();
+
+    // Define the SPI pins and create the SPI interface
+    let sck = io.pins.gpio12;
+    let miso = io.pins.gpio11;
+    let mosi = io.pins.gpio13;
+    let cs = io.pins.gpio10;
+    let spi = Spi::new(
+        peripherals.SPI2,
+        sck,
+        mosi,
+        miso,
+        cs,
+        100_u32.kHz(),
+        SpiMode::Mode0,
+        &mut system.peripheral_clock_control,
+        &clocks,
+    );
+
+    // Define the display interface with no chip select
+    let di = SPIInterfaceNoCS::new(spi, dc);
+
+    // Define the display drom the display interface and initialize it
+    let mut display = Builder::ili9486_rgb565(di)
+        .init(&mut delay, Some(rst))
+        .unwrap();
+
+    // Make the display all black
+    display.clear(Rgb565::BLACK).unwrap();
+
+    // Turn SPRITE into an array of pixels in the right colorspace
+    let sprite: [Rgb565; 256] = SPRITE.map(|(r, g, b)| Rgb565::new(r, g, b));
+
+    // Draw the 16*16 sprite on the top left corner of the screen
+    display.set_pixels(0, 0, 15, 15, sprite).unwrap();
+
+    loop {
+        // Do nothing
+    }
+}
+
+// Contains all the pixel colors for a 16*16 sprite as a 256 value array of (r, g, b) values
+const SPRITE: [(u8, u8, u8); 256] = [
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (0xff, 0, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 69, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (255, 255, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (124, 252, 0),
+    (0, 191, 255),
+    (0, 191, 255),
+    (0, 191, 255),
+    (0, 191, 255),
+    (0, 191, 255),
+    (0, 191, 255),
+    (0, 191, 255),
+    (0, 191, 255),
+    (0, 191, 255),
+    (0, 191, 255),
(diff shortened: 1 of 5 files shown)


examples/spi-ili9486-esp32-c3/src/main.rs
```diff
@@ -0,0 +1,364 @@
...
+    Delay, Rtc, IO,
+};
+/* -------------------------- */
+
+use embedded_graphics::{
+    // Provides the necessary functions to draw on the display
+    draw_target::DrawTarget,
+    // Provides colors from the Rgb565 color space
+    pixelcolor::Rgb565,
+    prelude::RgbColor,
```
rfuest, line 21:
```suggestion
    prelude::*,
```
or if you don't want to use the prelude:
```suggestion
    pixelcolor::RgbColor,
```

examples/spi-ili9486-esp32-c3/src/main.rs
```diff
@@ -0,0 +1,364 @@
...
+        .unwrap();
+
+    // Make the display all black
+    display.clear(Rgb565::BLACK).unwrap();
+
+    // Turn SPRITE into an array of pixels in the right colorspace
+    let sprite: [Rgb565; 256] = SPRITE.map(|(r, g, b)| Rgb565::new(r, g, b));
+
+    // Draw the 16*16 sprite on the top left corner of the screen
+    display.set_pixels(0, 0, 15, 15, sprite).unwrap();
```
rfuest, line 99: This is pretty inefficient, because the `sprite` array needs to be build in RAM. I'm biased and would simply use embedded-graphics to draw an image, but if you want to demonstrate `set_pixels` I would suggest to change the `SPRITE` constant to be `SPRITE: [Rgb565; 256]` instead.

In addition to that the `SPRITE` constant seems to use values between 0 and 255 while `Rgb565` expects 5 or 6 bit values depending on the channel.

rfuest (changes requested):

examples/spi-ili9486-esp32-c3/src/main.rs
```diff
@@ -0,0 +1,364 @@
...
+        100_u32.kHz(),
+        SpiMode::Mode0,
+        &mut system.peripheral_clock_control,
+        &clocks,
+    );
+
+    // Define the display interface with no chip select
+    let di = SPIInterfaceNoCS::new(spi, dc);
+
+    // Define the display drom the display interface and initialize it
```
rfuest, line 87:
```suggestion
    // Define the display from the display interface and initialize it
```

src/lib.rs
```diff
@@ -179,6 +179,23 @@ where
     /// * `ey` - y coordinate end
     /// * `colors` - anything that can provide `IntoIterator<Item = u16>` to iterate over pixel data
     ///
+    /// # Example
```
rfuest, line 182: If you want to improve the docs for `set_pixel` further:

One thing that is important to document IMO is the behavior of `set_pixels` if the `colors` iterator contains more pixels than required by the rectangle bounds. If I remember correctly there is no bounds checking and the drawing operation would just wrap back to the top left of the rectangle.

This is different to `embedded-graphics`'s similar [`fill_contiguous` method](https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTarget.html#method.fill_contiguous) and should be highlighted. Because this makes it impossible to just use something like `display.set_pixel(0, 0, 10, 10, core::iter::repeat(Rgb565::RED))`, which works in e-g.

examples/spi-ili9486-esp32-c3/src/main.rs
Slushee-a: I did want to demonstrate how `map` can be used to turn a regular array into an array of RgbXYZ colors, but it does make more sense to just use `embedded-graphics` I will change this.

rfuest: If you want to make this an example of using `set_pixels` I'm OK with that, but the example should make it clear that there are other options available. For more memory constrained targets the array in RAM could be a problem for larger images.

You don't need an array, because `set_pixels` takes an iter: `display.set_pixels(0, 0, 15, 15, SPRITE.map(|(r, g, b)| Rgb565::new(r, g, b))).unwrap();` should also work, without the memory overhead.

Slushee-a:
# Changes included on the latest commit:
 - Removed `map` example from [`set_pixels`](https://docs.rs/mipidsi/latest/mipidsi/struct.Display.html#method.set_pixels), swapped directly with an Rgb666 array instead
 - Removed `map` + `set_pixels` segment from `spi-ili9486-esp32-c3` example in favor of [`fill_contiguous`](https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTarget.html#method.fill_contiguous) with an `Rgb565` array
 - Added an example drawing of a basic smiley face on the `spi-ili9486-esp32-c3` example
 - Fix typo
 - Fix dependency usage

src/lib.rs
Slushee-a: > This is different to embedded-graphics's similar [`fill_contiguous` method](https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTarget.html#method.fill_contiguous) and should be highlighted

I added a `# Warning` [**here »**](https://github.com/almindor/mipidsi/pull/66/commits/e611757fa4155a66efd5724e5a1d50896a0f739f#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R192) mentioning that using [`fill_contiguous`](https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTarget.html#method.fill_contiguous) instead is recommended. Does this sound good or should I highlight that in another way?

examples/spi-ili9486-esp32-c3/src/main.rs
Slushee-a: I opted to use `const SPRITE: [Rgb565; 256] = [Rgb565::RED, Rgb565::ORANGE ...` instead with [`fill_contiguous`](https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTarget.html#method.fill_contiguous). If this is okay you can resolve the conversation. Otherwise do suggest something

src/lib.rs
rfuest: I would move it to above the argument list and rephrase it a bit. It's quite important to know the behavior of and IMO it's harder to notice the warning if it is below the example.

Instead of a warning I would just describe the behavior of this method. A rough draft of a description:
> No bounds checking is performed on the `colors` iterator, which can lead to unexpected behavior if the iterator returns more or less color values than the number of pixels in the given update region. Use [`fill_contiguous`](https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTarget.html#method.fill_contiguous) from `embedded-graphics` as an alternative with bounds checking.

rfuest (changes requested):

Slushee-a:
# Changes included on the latest commit:
 - Switched away from `Rectangle::with_corners` in favor of `Rectangle::new`
 - Move smiley face drawing from main to a separate function
 - Remove outdated comment (Sorry!)
 - Move and rephrase warning on [`set_pixels()`](https://docs.rs/mipidsi/latest/mipidsi/struct.Display.html#method.set_pixels)

examples/spi-ili9486-esp32-c3/src/main.rs
```diff
@@ -0,0 +1,399 @@
...
+        .init(&mut delay, Some(rst))
+        .unwrap();
+
+    let sprite_rectangle = Rectangle::with_corners(Point::new(0, 0), Point::new(15, 15));
+
+    // make the display all black
+    display.clear(Rgb565::BLACK).unwrap();
+
+    // draw the 16*16 sprite on the top left corner of the screen
+    display.fill_contiguous(&sprite_rectangle, SPRITE).unwrap();
```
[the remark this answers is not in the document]
Slushee-a: I think it is best to make the examples as 'correct' on the usage as possible, since they're made to demonstrate how things are intended to be done. 
Maybe this example can be left as a 'possibility', but I think an example using `ImageRaw` and `Image` is important, since drawing custom sprites to the screen is probably a common use for the crate.
To be fair though, this is done by `embedded-graphics` so maybe it's out of scope.

src/lib.rs
almindor: One note here is that `set_pixels` is essentially a driver primitive. It's not meant to be used directly and is exposed so frameworks other than `embedded-graphics` can be used.

almindor (approved):
LGTM but I'll wait for @rfuest to +1 as well

examples/spi-ili9486-esp32-c3/src/main.rs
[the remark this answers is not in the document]
rfuest: I would either remove the image drawing from this example or change it to something that is easy to use. Drawing pixels from an array that is manually added to the code isn't very common IMO.

Using the `tinybmp` instead of `ImageRaw` would make it much easier for a user to generate the necessary image data, because they can just export a BMP file from an image editor like Gimp. The BMP format does also natively support RGB565 colors, which makes it a good fit for the used display. See the `tinybmp` for an example how to use it: https://github.com/embedded-graphics/tinybmp#draw-a-bmp-image-to-an-embedded-graphics-draw-target

Regardless of whether we decide to remove the image drawing or not, a link to the `embedded-graphics` example would be a good idea to get an overview of the possible drawing operations: https://docs.rs/embedded-graphics/latest/embedded_graphics/examples/index.html

src/lib.rs
```diff
@@ -168,8 +168,12 @@ where
         Ok(())
     }
 
+    /// Sets pixel colors in any given rectangle bounds. Does not perform any bounds checking on the input color iterator.
+    /// It's not meat to be used directly, it's exposed for frameworks to be able to use it.
```
rfuest, line 172: The first line of doc comments should be a summary (https://doc.rust-lang.org/1.0.0/style/style/comments.html#summary-line). 

Here is another attempt to make this easier to understand:
```suggestion
    /// Sets pixel colors in a rectangular region.
    ///
    /// The color values from the `colors` iterator will be drawn to the given region starting
    /// at the top left corner and continuing, row first, to the bottom right corner. No bounds
    /// checking is performed on the `colors` iterator and drawing will wrap around if the
    /// iterator returns more color values than the number of pixels in the given region.
    ///
    /// This is a low level function, which isn't intended to be used in regular user code.
    /// Consider to use the drawing functions provided by the `embedded-graphics` crate instead.
```
GitHub didn't allow me to add a suggestion which includes all lines until the "Arguments" heading, but this is meant to replace the entire first part of the comment (up to the "Arguments" heading).

I did leave out the detail about using `fill_contiguous`, because that is also a low level function which shouldn't be used directly.

rfuest (changes requested):

Slushee-a:
I do agree with the fact that the `fill_contiguous` part of the example should be removed, since drawing basic shapes is already enough for the example. Adding a link to the `embedded-graphics` examples is a great idea

src/lib.rs
```diff
@@ -169,7 +169,16 @@ where
...
+    /// Sets pixel colors in a rectangular region.
+    ///
+    /// The color values from the `colors` iterator will be drawn to the given region starting
+    /// at the top left corner and continuing, row first, to the bottom right corner. No bounds
+    /// checking is performed on the `colors` iterator and drawing will wrap around if the
+    /// iterator returns more color values than the number of pixels in the given region.
+    ///
+    /// This is a low level function, which isn't intended to be used in regular user code.
+    /// Consider using the [`fill_contiguous`](https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTarget.html#method.fill_contiguous)
+    /// function from the `embedded-graphics` crate as an instead.
```
rfuest, line 181: Something is missing in this sentence:
```suggestion
    /// function from the `embedded-graphics` crate as an alternative instead.
```

rfuest (approved):
Looks good, except for a missing word in one of the comments. Thanks for your efforts to improve the examples and docs. It's sometimes hard to know what is missing if you are already very familiar with the create.

rfuest:
> Adding a link to the `embedded-graphics` examples is a great idea. Should it be added to the docs homepage or to the README?

Both locations would make sense, some users might learn about the crate via the README while others read the docs first. The README already contains a reference to e-g with this sentence: "embedded-graphics-core is used to provide the drawing API.", but that isn't very useful for users of the crate IMO. The distinction between `embedded-graphics-core` and `embedded-graphics` should be considered an implementation detail and not something that is important for users. A short description what e-g is and links to the relevant docs would be more useful than a link to e-g-core.

src/lib.rs
Slushee-a: Oops! I don't know how that slipped through. Thank you for all the patience and help!

> It's sometimes hard to know what is missing if you are already very familiar with the create.

And coming from the outside, it's difficult to tell how something is meant to be used without clear instructions - That's why I was so off with `set_pixels` and why I wanted to add some examples :)

PR #70, opened 2023-08-13
Add spi st7789 rpi zero w example

martinschaer:
Tested with [Display HAT Mini by Pomoroni](https://shop.pimoroni.com/products/display-hat-mini?variant=39496084717651), and added instructions to cross-compile from Mac OS.


mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+        .with_display_size(w, h)
+        // .with_orientation(mipidsi::Orientation::Landscape(false))
+        .with_invert_colors(mipidsi::ColorInversion::Inverted)
+        .init(&mut delay, None::<OutputPin>)
+        .unwrap();
+
+    // Turn on backlight
+    backlight.set_low();
+    sleep(Duration::from_millis(150));
+    backlight.set_high();
```
rfuest, line 43: Why do you need to sleep for 150ms to turn the backlight on?

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    // Clear the display initially
+    let style = PrimitiveStyleBuilder::new()
+        .fill_color(Rgb565::BLACK)
+        .build();
+    Rectangle::new(Point::new(0, 0), Size::new(w.into(), h.into()))
+        .into_styled(style)
+        .draw(&mut display)
+        .unwrap();
```
rfuest, line 52:
```suggestion
display.clear(Rgb565::BLACK).unwrap();
```

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    sleep(Duration::from_millis(150));
+
+    loop {
+        // Fill the display with red
```
rfuest, line 56: The drawing operations in this loop could also be replaced with `display.clear`, but I would suggest to change the example to draw anything else and not only fill the display with different colors.

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+        sleep(Duration::from_millis(500));
+
+        // Clear the display
+        let style = PrimitiveStyleBuilder::new()
+            .fill_color(Rgb565::BLUE)
+            .build();
+        Rectangle::new(Point::new(0, 0), Size::new(w.into(), h.into()))
+            .into_styled(style)
+            .draw(&mut display)
+            .unwrap();
```
rfuest, line 73: There is a way to get a rectangle for the display without building a `Rectangle` manually:
```suggestion
        display
            .bounding_box()
            .into_styled(style)
            .draw(&mut display)
            .unwrap();
```

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
martinschaer: It was just a test. Removed in 450a3fd70a539c49f26c44c7bf1b97ae6a8da185

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
martinschaer: @rfuest thanks!

almindor:
Thanks for the example! Seems there are some issues with the CI/builds.

Also as a side note I created https://github.com/almindor/mipidsi-examples as a separate repo for just example. It's been somewhat underused and not properly referenced, but I'm wondering if we should move things there and create some sort of structure of MCU-display pairs.

martinschaer:
@almindor I pushed some fixes for the CI, hopefully the checks pass now

I'll take a look at the other repo!

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        // Fill the display with red
+        display.clear(colors[color_index]).unwrap();
+        color_index = (color_index + 1) % colors.len();
+
+        // Draw text
+        Text::new(text, Point::new(text_x, H / 2), character_style)
+            .draw(&mut display)
+            .unwrap();
+
```
rfuest, line 74: You can use the return value of `Text::draw` to check if the text has reached the left edge of the display and is no longer visible, without needing to calculate `text_w` manually: 
```suggestion
        let right = Text::new(text, Point::new(text_x, H / 2), character_style)
            .draw(&mut display)
            .unwrap();
            
        text_x = if right.x <= 0 { W } else { text_x - 6 };
```

rfuest (changes requested):
Most comments are just suggestions for possible improvements, but the pinout issue should be addressed before this gets merged.

mipidsi/examples/spi-st7789-rpi-zero-w/README.md
```diff
@@ -0,0 +1,30 @@
...
+Tested with [Display HAT Mini by Pomoroni](https://shop.pimoroni.com/products/display-hat-mini?variant=39496084717651).
+
+## Build, strip, copy, and run
+
+### Mac OS
+
+Prerequisite: musl-cross (Homebrew users: `brew install filosottile/musl-cross/musl-cross`)
+
+```bash
+# build for rpi zero 2 w
```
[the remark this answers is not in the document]
martinschaer: Good catch. The comment is wrong, I meant `rpi zero w` which is the one I'm testing on. Version 2 is ARMv8-A (64/32-bit), while version 1 is ARMv6Z (32-bit). Sadly I can't test version 2, but from what I've read [here](https://github.com/japaric/rust-cross) it may work but won't include optimizations, due to be using `arm` in the triple instead of a more specific one like `armv6`. I'm going to fix the comment

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+use mipidsi::Builder;
+use rppal::gpio::{Gpio, OutputPin};
+use rppal::hal::Delay;
+use rppal::spi::{Bus, Mode, SlaveSelect, Spi};
+use std::process::ExitCode;
+use std::thread::sleep;
+use std::time::Duration;
+
+const SPI_DC: u8 = 9;
+const SPI_CS: u8 = 1;
```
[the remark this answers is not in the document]
rfuest: My guess is that there is a mistake in the Pyhton code and it works by accident, because the CS pin is low all the time. To test this theory I would just leave the code as it is and configure GPIO 7 as an high output. If it still works CS is connected to GPIO 1 and if not it is connected to GPIO 7.

> Why would one use pin 1 (3v3)? Is it a mistake in the python code?

Keep in mind that the GPIO pin numbers are different from the physical pin numbers on the 40 pin connector. GPIO 1 is connected to pin 28 (https://pinout.xyz/pinout/pin28_gpio1/#), which is also very close to GPIO 7.

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    loop {
+        // Text scroll
+        text_x = (text_x - 6) % (W + text_w);
+        if text_x < -text_w {
+            text_x = W;
+        }
+
+        // Fill the display with red
+        display.clear(colors[color_index]).unwrap();
+        color_index = (color_index + 1) % colors.len();
```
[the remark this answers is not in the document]
martinschaer: I ended up not using the iterator to be able to alternate the colors every 8 frames, but thanks for the suggestion, I didn't know cycle iterators

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+        if button_x.is_low() {
+            text_y -= char_h;
+        }
+        // Y: move text down
+        if button_y.is_low() {
+            text_y += char_h;
+        }
+        // A: change led color
+        if button_a.is_low() {
+            led_flags = (led_flags + 1) % 7;
```
rfuest, line 108: The modulo is off by one:
```suggestion
            led_flags = (led_flags + 1) % 8;
            //or: led_flags = (led_flags + 1) & 0b111;
```

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+    let mut counter = 0;
+    loop {
+        let elapsed = last.elapsed().as_secs_f64();
+        if elapsed < 0.125 {
+            continue;
+        }
+        last = std::time::Instant::now();
+        counter += 1;
+
+        // X: move text up
```
rfuest, line 98: A brief description of what the example demonstrates and a summary of the button functions at the top of the example would be helpful.

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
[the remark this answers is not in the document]
martinschaer: You were right once again, thanks! Updated the readme with the pre-requisite

rfuest (approved):
LGTM

almindor (approved):

PR #75, opened 2023-10-03
feat: Added sleep/wake methods to display

xgroleau:
I've added the `sleep` and `wake` method to the `Display` struct. Note that I have not bothered with `idle` since it only allows 8bit colors, but we can add it if you prefer. 

It saves about 20mW (5mA at 4V) to just put the display to sleep when the display is off.

Here is the reference of what the different settings of the modes do (for the st7789 anyway)
```
1. Normal Mode On (full display), Idle Mode Off, Sleep Out.
In this mode, the display is able to show maximum 262,144 colors.

2. Partial Mode On, Idle Mode Off, Sleep Out.
In this mode part of the display is used with maximum 262,144 colors.

3. Normal Mode On (full display), Idle Mode On, Sleep Out.
In this mode, the full display area is used but with 8 colors.

4. Partial Mode On, Idle Mode On, Sleep Out.
In this mode, part of the display is used but with 8 colors.

5. Sleep In Mode
In this mode, the DC: DC converter, internal oscillator and panel driver circuit are stopped. Only the MCU
```

Pretty simple PR since it was easy to add to the codebase but feel free to add any suggestions

### DIFF
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -253,7 +253,7 @@ where
         self.dcs
             .write_command(dcs::SetTearingEffect(tearing_effect))
     }
-  
+
     ///
     /// Puts the display to sleep, reducing power consumption.
     /// Need to call [Self::wake] before issuing other commands
@@ -261,7 +261,7 @@ where
     pub fn sleep(&mut self) -> Result<(), Error> {
         self.dcs.write_command(dcs::EnterSleepMode)
     }
-  
+
     ///
     /// Wakes the display after it's been set to sleep via [Self::sleep]
     ///
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -80,6 +80,7 @@ use dcs::Dcs;
 use display_interface::WriteOnlyDataCommand;
 
 pub mod error;
+use embedded_hal::blocking::delay::DelayUs;
 use embedded_hal::digital::v2::OutputPin;
 pub use error::Error;
 
@@ -258,14 +259,18 @@ where
     /// Puts the display to sleep, reducing power consumption.
     /// Need to call [Self::wake] before issuing other commands
     ///
-    pub fn sleep(&mut self) -> Result<(), Error> {
-        self.dcs.write_command(dcs::EnterSleepMode)
+    pub fn sleep<D: DelayUs<u32>>(&mut self, delay: &mut D) -> Result<(), Error> {
+        self.dcs.write_command(dcs::EnterSleepMode)?;
+        delay.delay_us(120_000); // All supported models requires a 120ms delay
+        Ok(())
     }
 
     ///
     /// Wakes the display after it's been set to sleep via [Self::sleep]
     ///
-    pub fn wake(&mut self) -> Result<(), Error> {
-        self.dcs.write_command(dcs::ExitSleepMode)
+    pub fn wake<D: DelayUs<u32>>(&mut self, delay: &mut D) -> Result<(), Error> {
+        self.dcs.write_command(dcs::ExitSleepMode)?;
+        delay.delay_us(120_000); // ST77xx have the highest wait delay of 120ms
+        Ok(())
     }
 }
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -261,7 +261,8 @@ where
     ///
     pub fn sleep<D: DelayUs<u32>>(&mut self, delay: &mut D) -> Result<(), Error> {
         self.dcs.write_command(dcs::EnterSleepMode)?;
-        delay.delay_us(120_000); // All supported models requires a 120ms delay
+        // All supported models requires a 120ms delay before issuing other commands
+        delay.delay_us(120_000);
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


mipidsi/src/lib.rs
```diff
@@ -253,4 +253,19 @@ where
         self.dcs
             .write_command(dcs::SetTearingEffect(tearing_effect))
     }
+
+    ///
+    /// Puts the display to sleep, reducing power consumption.
+    /// Need to call [Self::wake] before issuing other commands
+    ///
+    pub fn sleep(&mut self) -> Result<(), Error> {
```
almindor, line 261: Perhaps I'm overcomplicating things but I wonder if we should make this a `self` consuming method that returns some sort of `SleepingDisplay<Display<M...>>` that will only return the `Display<M...>` back by issueing `wake(self) -> Display<M...>` to it.

@rfuest what do you think, perhaps i'm just falling into the Rust overabstraction trap tho.

xgroleau: I've dabbled with typestates for other drivers and from what I experienced it becomes a bit of a pain when they are fallible since you need to return the previous state with the error. It's still possible though, just the error case is annoying to handle.

Also not really possible to do async cancellation since it takes ownership if you eventually want the same api for async.

xgroleau: Wait, I thought you were proposing a typestate, but a wrapper over it is actually not a bad idea, just annoying for the owner

```rs
pub enum DisplayDriverEnum {
    ASleep(SleepingDisplay<Display<M...>>),
    Awake(Display<M...>)
}
pub struct UI {
   display: DisplayDriverEnum,
}

```

rfuest: > @rfuest what do you think, perhaps i'm just falling into the Rust overabstraction trap tho.

I agree with @xgroleau that adding a typestate of wrapper type will become a hassle for the user of the library and isn't worth it in this case IMO.

rfuest:
Some comments:
- If we don't use a wrapper type or typestates to track the sleep state we should have a `sleep_state() -> SleepState` (or similar) getter to allow the user to get the current sleep state.
- There are some timing requirements when entering/exiting the sleep mode, how do we make sure these aren't violated?
- Not all commands are valid in sleep mode. I'm not sure if we use any commands that aren't valid in sleep mode at the moment, but if we do they should return a special error to notify the user that the operation isn't supported in sleep mode.

almindor:
> Some comments:
> 
> * If we don't use a wrapper type or typestates to track the sleep state we should have a `sleep_state() -> SleepState` (or similar) getter to allow the user to get the current sleep state.
> * There are some timing requirements when entering/exiting the sleep mode, how do we make sure these aren't violated?
> * Not all commands are valid in sleep mode. I'm not sure if we use any commands that aren't valid in sleep mode at the moment, but if we do they should return a special error to notify the user that the operation isn't supported in sleep mode.

We should def. have a state getter, but it means an additional variable on the display level.

All good concerns. When it comes to timing we'll probably have to give the driver a `&mut DelayUs` to the command calls. The timings can *probably* be unified accross models.

AFAIK we don't use invalid commands in sleep mode, since the only commands outside of init we use are windowing/drawing.

xgroleau:
> All good concerns. When it comes to timing we'll probably have to give the driver a `&mut DelayUs` to the command calls. The timings can _probably_ be unified accross models.

I could see a method to the `DcsCommand` trait that where we would pass the `&mut DelayUs`? We could then specify the delay in the `dcs_basic_command` macro and the DCS would handle the delay.

Ex:
```rs
pub trait DcsCommand {
    /// Returns the instruction code.
    fn instruction(&self) -> u8;

    /// Fills the given buffer with the command parameters.
    fn fill_params_buf(&self, buffer: &mut [u8]) -> Result<usize, Error>;
    
    /// Delay to wait after the command
    fn post_command_delay<D: DelayUs>(&self, delay: &mut D); // I think DelayUs is now infaillable, so we might not need a result, or we return the number of ms
    /// Or instead use a const
    const POST_COMMAND_DELAY: u32;
}
```

Then if we go this way, should the `Display` or `DCS` struct take ownership of the `DelayUs`?

rfuest:
> I could see a method to the `DcsCommand` trait that where we would pass the `&mut DelayUs`

`DcsCommand` isn't the correct location to add a delay. `DcsCommand` is only used to represent DCS commands and convert them to the corresponding byte sequence. It doesn't do any communication or handle the higher level requirements a command might have, like delays.

> Then if we go this way, should the `Display` or `DCS` struct take ownership of the `DelayUs`?

At the moment it should be enough if the `sleep` and `wake` methods take a borrowed `DelayUs`. To reduce unnecessary delays we would need to be able to measure the time between the `wake` call and the previous `sleep` call, which I don't think is possible with just `embedded_hal`. So we will need to assume the worst case and add pretty long delays to `sleep` and `wake` instead.

almindor (changes requested):
So we'll need to add the time delay handling. I'd just make it a `fn sleep(&mut self, delay: &mut dyn DelayUs) -> Result<(), Error>` kind of thing and choose the largest delay we use for sleep waits in all the models for now.

Same for wakeup.

almindor (changes requested):
Almost there, I forgot could you please add a CHANGELOG entry as well? Sorry I should've caught this in the last review. 

Thank you!

xgroleau:
From the doc, all the currently supported display requires a 120ms delay after sleep. For wake, the ST line requires the highest wait before issuing commands with 120ms. 

Though the user _could_ implement another `Model` himself  in his user code  since it's public. The timing might be different but there's no wait to control that really.

almindor (approved):
Thank you!

rfuest:
Didn't we decide to add a sleep state getter?

almindor:
> Didn't we decide to add a sleep state getter?

:man_facepalming: 

Sorry, I'll add an issue for it so I don't forget to add it.

PR #78, opened 2023-10-06
add Display::is_sleeping fixes #77

### LINKED ISSUE #77: Add a sleep state + getter
As discussed in #75

almindor:
Adds a preliminary `is_sleeping` method. I'm not too happy about the state machine logic here though. I think we can merge this for the next version but I want to refactor things so that we can have a workable state machine for things like this that can properly tie into the Model inits etc.

### DIFF
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -259,7 +259,6 @@ where
 
     ///
     /// Returns `true` if display is currently set to sleep.
-    /// NOTE: value before call to `init` is considered undefined.
     ///
     pub fn is_sleeping<D: DelayUs<u32>>(&self) -> bool {
         self.sleeping
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -119,6 +119,8 @@ where
     options: ModelOptions,
     // Current MADCTL value copy for runtime updates
     madctl: dcs::SetAddressMode,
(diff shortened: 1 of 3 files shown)


almindor:
@rfuest what do you think? I think we'll need to add some sort of state machine that can work well accross model and display itself.

In this case specifically I'm unhappy about how the sleeping state is not properly tied to init.

rfuest:
> In this case specifically I'm unhappy about how the sleeping state is not properly tied to init.

I'm not sure I understand what you mean. Would you prefer that the initial state can be set by `Model::init` instead of `Display::init`? For the sleeping state I don't think this is necessary, because `Model::init` is required to initialize the display into a non sleeping state. I also can't think of any other state that shouldn't be initialized to the same value by all `Model::init` implementations. Do you have a use case in mind?

mipidsi/src/lib.rs
```diff
@@ -255,6 +257,14 @@ where
             .write_command(dcs::SetTearingEffect(tearing_effect))
     }
 
+    ///
+    /// Returns `true` if display is currently set to sleep.
+    /// NOTE: value before call to `init` is considered undefined.
```
rfuest, line 262: Isn't it impossible to get an instance of `Display` without calling `Builder::init`?

almindor: true that, removed

rfuest (approved):
LGTM

PR #79, opened 2023-10-07
Add test image

rfuest:
This PR adds a simple test image. The main reason of adding a test image is to help users identify issues with the color settings or to find the correct orientation for their display/application. But it can also be useful to just get some initial output when you want to try a new display and don't have any drawing code implemented yet.

testimage

The troubleshooting docs will need to be updated to point to the `TestImage` and to provide some guidance to diagnose common issues, but I think that should be handled in another PR.

### DIFF
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -97,6 +97,9 @@ use models::Model;
 
 mod graphics;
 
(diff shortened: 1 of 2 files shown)


almindor (approved):
Great idea!

Issue #53, opened 2023-02-19
Remove `window_offset_handler`


rfuest:
Having to implement a `window_offset_handler` to use different display sizes shouldn't be necessary and it makes using different display sizes harder.

Only three parameters are required to calculate the  correct transformation from logical coordinates to the memory locations in the display framebuffer:
- The maximum supported display size for the controller (aka the framebuffer size)
- The size of the display
- The offset in the default orientation

Let's assume an 32x32 pixel display with a ILI9341 controller, that supports max. 240x320 pixel. (Nobody would use this combination, but it's just an example). For the example I'll only look at the X axis, but the Y axis will work the same way.

If the offset in the default orientation is 10 pixel the x coordinates in the framebuffer will be in the range `10..10+32 = 10..42`. If the X axis gets flipped by using MADCTL the display will now be accessible using the range `240 - (10 + 32)..240 - 10 = 198..230`.

I think a good way to implement this is to make a better distinction between displays models and display driver ICs inside the crate. The maximum display size is a physical property of each controller IC type and should not be modifiable in `ModelOptions`.

Roughly something like this:
```rust
trait DisplayController {
    const MAX_SIZE: (usize, usize);
    ... // methods like init
}

struct ST7789;

impl DisplayController for ST7789 {
    const MAX_SIZE: (usize, usize) = (240, 320);
}

trait DisplayModel {
    type Controller: DisplayController;
    const SIZE: (usize, usize);
    const OFFSET: (usize, usize) = (
        (Self::Controller::MAX_SIZE.0 - Self::SIZE.0) / 2,
        (Self::Controller::MAX_SIZE.1 - Self::SIZE.1) / 2,
    );
    const COLOR_ORDER: ColorOrder = ColorOrder::Rgb;
    const INVERTED: ColorInversion = ColorInversion::Normal;
}

struct Pico1;

impl DisplayModel for Pico1 {
    type Controller = ST7789;
    const SIZE: (usize, usize) = (135, 240);
    // The default OFFSET could be overridden, but is correct in this case.
    const INVERTED: ColorInversion = ColorInversion::Inverted;
}
```

The builder would then either take a display, like `Pico1`, which already contains all necessary settings. Or you could define a new display based on the used controller IC:
```rust
let mut display = Builder::with_display(Pico1, di)
    .init(&mud delay, Some(rst))?;
// or
let mut display = Builder::with_controller(ST7789, di)
    .with_display_size(135, 240)
    ...
[...]

PR #80, opened 2023-10-07
Add new Orientation type and remove window_offset_handler

rfuest:
This PR adds the `Orientation` and `Rotation` types, which were originally intended to be added to the `display-driver-hal` crate. I don't think that crate will be released any time soon, because there didn't seem to be any interest from other driver authors. But I think the new types are easier to use and we can still extract them into another crate later.

There was at least one orientation which didn't work correctly in the previous implementation, which should be fixed now. I don't remember the details, but I think one orientation was mirrored, even without setting the boolean mirrored flag.

I also decided to remove the `window_offset_handler` as a part of these changes, because I had to modify the code anyway. Instead of a handler there are now two settings, one for the display size and one for the offset. The framebuffer size is constant for every controller and is now a new `const` in the `Model` trait.

The PR is marked as a draft at the moment, because I still need to tweak/test a few things:
- [x] Fix `SetScrollArea`
- [ ] Set default offset to `framebuffer_size - display_size / 2`
- [ ] Remove `default_options`? This would require separate struct/traits for models and controller, like I suggested in #53.

### DIFF
--- a/mipidsi/src/dcs/set_scroll_area.rs
+++ b/mipidsi/src/dcs/set_scroll_area.rs
@@ -47,7 +47,8 @@ impl From<&ModelOptions> for SetScrollArea {
     fn from(options: &ModelOptions) -> Self {
         Self {
             tfa: 0,
-            vsa: options.framebuffer_size_max(),
+            //vsa: options.framebuffer_size_max(),
+            vsa: 0, //TODO: fix
             bfa: 0,
         }
     }
--- a/mipidsi/src/dcs/set_scroll_area.rs
+++ b/mipidsi/src/dcs/set_scroll_area.rs
@@ -1,7 +1,6 @@
 //! Module for the VSCRDEF visual scroll definition instruction constructors
 
 use crate::error::Error;
-use crate::options::ModelOptions;
 
 use super::DcsCommand;
 
@@ -40,17 +39,6 @@ impl DcsCommand for SetScrollArea {
         buffer[5] = bfa_bytes[1];
 
         Ok(6)
-    }
-}
-
-impl From<&ModelOptions> for SetScrollArea {
-    fn from(options: &ModelOptions) -> Self {
-        Self {
-            tfa: 0,
-            //vsa: options.framebuffer_size_max(),
-            vsa: 0, //TODO: fix
-            bfa: 0,
-        }
     }
 }
 
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -86,6 +86,7 @@ use embedded_hal::blocking::delay::DelayUs;
 use embedded_hal::digital::v2::OutputPin;
 
 pub mod options;
+use options::MemoryMapping;
 
 mod builder;
 pub use builder::Builder;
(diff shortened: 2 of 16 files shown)


rfuest:
If you want to try the new offset code yourself, here is how I test the code without having a lot of smaller displays with different offsets:
```rust
// Initialize the full screen display
let mut tft = Builder::ili9341_rgb565(tft_di)
    .with_orientation(Orientation::default())
    .init(&mut delay, Some(interfaces.tft_reset))
    .ok()
    .unwrap();

let offset = Point::new(10, 40);
let size = Size::new(135, 240);

// Draw a border around the area which would be used
// by a smaller display with the given size/offset
tft.clear(Gray8::new(64).into()).unwrap();
Rectangle::new(offset, size)
    .into_styled(
        PrimitiveStyleBuilder::new()
            .stroke_color(Rgb565::MAGENTA)
            .stroke_width(1)
            .stroke_alignment(StrokeAlignment::Outside)
            .build(),
    )
    .draw(&mut tft)
    .unwrap();

// Reinitialize the display with the smaller size and offset
let (tft_di, _m, reset) = tft.release();
let mut tft = Builder::ili9341_rgb565(tft_di)
    .with_display_size(size.width as u16, size.height as u16)
    .with_display_offset(offset.x as u16, offset.y as u16)
    .init(&mut delay, reset)
    .ok()
    .unwrap();

// Draw the test image, which should always be displayed inside the magenta border,
// regardless of the orientation setting
let test_image = TestImage::new();
test_image.draw(&mut tft).unwrap();
```

almindor:
@rfuest how are we here? Do you still plan to finish this?

rfuest:
I didn't have time to finish this since I crated the PR and I'm not sure when I'll get to fix the remaining issues. But I think that removing the `window_offset_handler` is important to support different display sizes and this should be merged.

mipidsi/src/dcs/set_scroll_area.rs
```diff
@@ -43,16 +42,6 @@ impl DcsCommand for SetScrollArea {
...
-impl From<&ModelOptions> for SetScrollArea {
-    fn from(options: &ModelOptions) -> Self {
-        Self {
-            tfa: 0,
-            vsa: options.framebuffer_size_max(),
-            bfa: 0,
-        }
-    }
-}
-
```
rfuest, line 55: I've removed this because it wasn't correct and only used in one place anyway.

mipidsi/src/lib.rs
```diff
@@ -210,26 +211,49 @@ where
...
-    /// * `vsa` - Vertical scrolling area
-    /// * `bfa` - Bottom fixed area
+    /// The combined height of the fixed area must not larger than the
+    /// height of the framebuffer height in the default orientation.
     ///
-    pub fn set_scroll_region(&mut self, tfa: u16, vsa: u16, bfa: u16) -> Result<(), Error> {
-        let vscrdef = dcs::SetScrollArea::new(tfa, vsa, bfa);
+    /// After the scrolling region is defined the [`set_scroll_offset`] can be
+    /// used to scroll the display.
+    pub fn set_vertical_scroll_region(
```
rfuest, line 229: I've renamed this and added some more documentation to make it clear that this will only scroll vertically, regardless of the set orientation.

mipidsi/src/models/st7789.rs
```diff
@@ -49,7 +50,6 @@ impl Model for ST7789 {
         delay.delay_us(10_000);
 
         // set hw scroll area based on framebuffer size
-        dcs.write_command(SetScrollArea::from(options))?;
```
rfuest, line 52: This should be the default value, according to the datasheet.

rfuest:
I've marked the PR as ready for review. I would still like to get the last two remaining TODO items finished, but I think they are better handled in another PR and there might be some overlap with #81.

almindor (approved):
LGTM, thanks!

PR #90, opened 2024-01-12
make Display::dcs public, fixes #89

### LINKED ISSUE #89: Expose Display.dcs
Would it be feasible to expose the internal Dcs somehow?

i.e.
1. Make it public
2. Add a pass-through "write_raw" method to the Display struct
3. Implement a trait that exposes that functionality so that specific expansion methods could be created for each model outside of the shared abstraction.

Interested in sending custom commands to a ST7789.

Thanks for the lib...using it successfully (:

almindor:
Add `Display::dcs` `unsafe` method to allow sending custom commands. Fixes #89

### DIFF
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -112,8 +112,8 @@ where
     MODEL: Model,
     RST: OutputPin,
 {
-    /// DCS provider, can be used directly to send custom DCS commands to the device
-    pub dcs: Dcs<DI>,
+    // DCS provider
+    dcs: Dcs<DI>,
     // Model
     model: MODEL,
     // Reset pin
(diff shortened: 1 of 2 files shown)


rfuest:
I don't really like this change. If the user can send any command to the display it is possible to leave the controller in any state and possibly break further commands that are send using the regular API. I would prefer to only expose the DCS interface as an unsafe getter, if this function is really required and there is no way to add a higher level API to achieve the same function.

almindor:
@rfuest changed it into an `unsafe` getter method, lmk if this is ok.

mipidsi/src/lib.rs
```diff
@@ -289,4 +289,12 @@ where
         self.sleeping = false;
         Ok(())
     }
+
+    ///
+    /// Returns the DCS interface for sending raw commands.
+    /// *WARNING*: user is responsible for ensuring display state validity
+    ///
```
rfuest, line 296:
```suggestion
    /// Returns the DCS interface for sending raw commands.
    ///
    /// # Safety
    ///
    /// Sending raw commands to the controller can lead to undefined behaviour,
    /// because the rest of the code isn't aware of any state changes that were caused by sending raw commands.
    /// The user must ensure that the state of the controller isn't altered in a way that interferes with the normal
    /// operation of this crate.
```
Changed `*WARNING*` into `# Safety`, because of the API guidelines: https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure

PR #91, opened 2024-01-19
Make DCS command constructors const

bdolgov:
This will allow to define lists of commands, for example, in per display initialisation sequences:

```rust
pub const ILI9488_INIT_SEQ: &[InitOp] = &[
    InitOp::Cmd(&dcs::SetInvertMode(mipidsi::ColorInversion::Inverted)),
    InitOp::Cmd(&dcs::SetAddressMode::new(
        mipidsi::ColorOrder::Bgr,
        mipidsi::Orientation::Landscape(false),
        mipidsi::RefreshOrder {
            vertical: mipidsi::VerticalRefreshOrder::TopToBottom,
            horizontal: mipidsi::HorizontalRefreshOrder::LeftToRight,
        },
    )),
    ...
];
```


rfuest (approved):
LGTM. The only thing that's missing is an entry in the changelog.

almindor (changes requested):
LGTM too but please add the CHANGELOG entry

bdolgov:
Updated the changelog and amended the change into the same commit, please take a look.

Thanks for the quick review!

almindor (approved):

PR #97, opened 2024-01-22
do not re-exports sub-types from submodules, fixes #95

### LINKED ISSUE #95: Remove toplevel re-exports
I was just checking something in the docs and noticed the re-export section in the root of the create docs. Should we perhaps remove these re-exports?

I don't see the point of having the options available as e.g. `mipidsi::ColorOrder` and `mipidsi::options::ColorOrder`. Removing the re-export would keep the root namespace clean. And, from a docs point of view, the types wouldn't be any less discoverable, because the `options::*` re-export only links to the `options` module and doesn't give a full list of available types.

The `error` module is similar and I don't think there is a good reason for only re-exporting on of the two error types. Without the re-export all error types could be found in a single place.

almindor:
Removes the re-exports of submodule types

### DIFF
--- a/mipidsi/src/batch.rs
+++ b/mipidsi/src/batch.rs
@@ -1,7 +1,7 @@
 //! Original code from: [this repo](https://github.com/lupyuen/piet-embedded/blob/master/piet-embedded-graphics/src/batch.rs)
 //! Batch the pixels to be rendered into Pixel Rows and Pixel Blocks (contiguous Pixel Rows).
 //! This enables the pixels to be rendered efficiently as Pixel Blocks, which may be transmitted in a single Non-Blocking SPI request.
-use crate::{models::Model, Display, Error};
+use crate::{error::Error, models::Model, Display};
 use display_interface::WriteOnlyDataCommand;
 use embedded_graphics_core::prelude::*;
 use embedded_hal::digital::v2::OutputPin;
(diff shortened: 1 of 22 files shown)


almindor:
NOTE: I kept `Builder` as top level re-export because it's the entry point type + it's essentially the only thing in its module

mipidsi/src/batch.rs
```diff
@@ -1,7 +1,7 @@
 //! Original code from: [this repo](https://github.com/lupyuen/piet-embedded/blob/master/piet-embedded-graphics/src/batch.rs)
 //! Batch the pixels to be rendered into Pixel Rows and Pixel Blocks (contiguous Pixel Rows).
 //! This enables the pixels to be rendered efficiently as Pixel Blocks, which may be transmitted in a single Non-Blocking SPI request.
-use crate::{models::Model, Display, Error};
+use crate::{error::Error, models::Model, Display};
```
rfuest, line 4: At some point we might want to reformat all imports to be more consistent across the crate, but IIRC most rustfmt options for that aren't stabilized yet. If there was a fixed order of internal vs. external imports and the same grouping policy for all imports the diffs would be easier to read. But that's all outside the scope of this PR.

rfuest (changes requested):
This change needs to be mentioned in the changelog, because it is breaking.

And building the docs resulted in two warnings:
grafik

rfuest (approved):

PR #100, opened 2024-01-26
Fix DrawTarget::fill_contiguous

rfuest:
This PR fixes the implementation of the `fill_contiguous` method in the `DrawTarget` trait if `area` overlaps the edge of the framebuffer.

Here is an example that demonstrates the problem:

```rust
let logo = tinybmp::Bmp::from_slice(LOGO).unwrap();

for anchor in [
    AnchorPoint::TopLeft,
    AnchorPoint::TopRight,
    AnchorPoint::Center,
    AnchorPoint::BottomLeft,
    AnchorPoint::BottomRight,
] {
    Image::with_center(&logo, tft.bounding_box().anchor_point(anchor))
        .draw(&mut tft)
        .unwrap();
}
```

Current master version:
Before

After this PR is applied:
After


mipidsi/src/graphics.rs
```diff
@@ -50,24 +50,49 @@ where
...
-        if let Some(bottom_right) = area.bottom_right() {
-            let mut count = 0u32;
-            let max = area.size.width * area.size.height;
+        let intersection = area.intersection(&self.bounding_box());
+        let mut colors = colors.into_iter();
 
-            let mut colors = colors.into_iter().take_while(|_| {
-                count += 1;
-                count <= max
-            });
```
rfuest, line 60: Did the code use `take_while` instead of `take` to work on platforms where `usize` is 16 bit (like AVRs)? My code uses `nth` to skip pixels, which also takes a `usize` argument and has the same potential problems as `take`. I don't like to replace `nth` with a loop of `next` calls, which would make the code work for small `usize`s, but also reduce the performance on other platforms, if the `colors` iter has a specialized `nth` impl.

almindor: Yes, I believe so. It's been a while but I think the whole idea here was to avoid using usize due to 16bit platforms.

rfuest: OK, that's kind of annoying, but I'll try to find a solution that doesn't rely on `usize` > 16 bit.

rfuest:
I've updated the PR to also work on architectures with 16 bit `usize` by conditionally compiling a different implementation of `nth` and `take` in this case.

almindor (approved):

PR #104, opened 2024-01-27
Update `embedded-hal` to `1.0.0` and `display-interface` to `0.5.0`

markszente:
The examples have not been updated. I can probably do the ESP32-C3 one, and possibly the RP Pico as well. However, I cannot do the Zero W.


mipidsi/src/lib.rs
```diff
@@ -301,15 +301,15 @@ where
     ///
     /// Returns `true` if display is currently set to sleep.
     ///
-    pub fn is_sleeping<D: DelayUs<u32>>(&self) -> bool {
+    pub fn is_sleeping<D: DelayNs>(&self) -> bool {
```
rfuest, line 304: Not directly to this PR, but the type parameter should be removed:
```suggestion
    pub fn is_sleeping(&self) -> bool {
```

rfuest:
We will need to bump our MSRV to 1.75.0, because `display-interface` 0.5 does require at least that Rust version.

almindor:
@markszente thank you! you ninjad me on this :D Is it ready for review? I think it's good atm.

almindor:
Spoke too soon, the CI needs updating to MSRV 1.75 too

markszente:
@almindor I haven't touched the examples yet, so, I don't know. I hope to have some time for them later this week if they need updating. And yeah the CI too.

markszente:
Okay I guess the CI is taken care of

rfuest:
IMO we should merge this first and handle the examples in another PR. At least some of them were already broken before this PR. And perhaps we should also look into building the examples in CI, which is also outside the scope of this PR.

almindor:
@rfuest Agreed, merged this as-is we should fix up the examples (at least on paper, I know none of us have all the MCUs) in a separate PR before going for a release

rfuest:
I just realized a problem with building the examples in CI. The examples depend on the latest release of `mipidsi` and not on the current state of the `master` branch. Should we change the examples to use a `path` dependency instead?

PR #105, opened 2024-01-30
Update CI config and build RP pico example in CI

rfuest:
This PR updates the CI config to get rid of some deprecation messages.

I've also started to add CI jobs to build the examples. While we cannot run the examples this at least makes sure they remain buildable. This PR only adds a job to build the PR pico example, the other examples can be added in future PRs.


almindor (approved):
Thank you!

PR #111, opened 2024-02-02
Remove model specific constructors

rfuest:
This PR removes the model specific constructors from the `Builder` type. Instead one generic `new` constructor can now be used for all controller types. This shouldn't make the API harder to use and has other advantages, like reducing the amount of code required to implement a model. And this method also works for externally implemented `Model`s, which aren't able to add a new constructors to `Builder`.


almindor (changes requested):
LGTM, thank you, but could you please add an entry to the migration guide? This feels like it should be mentioned there for the next major version.

rfuest:
> LGTM, thank you, but could you please add an entry to the migration guide? This feels like it should be mentioned there for the next major version.

OK, done. Perhaps we should open an issue to remember to update the migration guide before the release. There are already other changes that should be mentioned in the guide, like the removed re-exports.

almindor (approved):

PR #115, opened 2024-02-14
fix cargo doc lint issues

### LINKED ISSUE #110: Run rustdoc in CI
Running `cargo doc` does currently show some warnings about broken links, but this isn't picked up by CI. We should add a job that runs `cargo doc` with the `RUSTDOCFLAGS` environment variable set to `-Dwarnings` to fail on doc issues.

almindor:
Simple docs lint fixes PR + adds docs build check to CI. Fixes #110

### DIFF
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -277,7 +277,7 @@ where
     /// Setting the vertical scroll offset shifts the vertical scroll region
     /// upwards by `offset` pixels.
     ///
-    /// Use [Self::set_vertical_scroll_region] to setup the scroll region, before
+    /// Use [`set_vertical_scroll_region`](Self::set_vertical_scroll_region) to setup the scroll region, before
     /// using this method.
     pub fn set_vertical_scroll_offset(&mut self, offset: u16) -> Result<(), Error> {
         let vscad = dcs::SetScrollStart::new(offset);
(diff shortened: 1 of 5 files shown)


mipidsi/src/lib.rs
```diff
@@ -250,7 +250,7 @@ where
     /// The combined height of the fixed area must not larger than the
     /// height of the framebuffer height in the default orientation.
     ///
-    /// After the scrolling region is defined the [`set_scroll_offset`] can be
+    /// After the scrolling region is defined the [Self::set_vertical_scroll_offset] can be
```
rfuest, line 253: It's common practice to not include `Self` when referring to methods in the same type, which makes the docs easier to read. Method names should also be inside a code span:
```suggestion
    /// After the scrolling region is defined the [`set_vertical_scroll_offset`](Self::set_vertical_scroll_offset) can be
```

Other parts of the docs also don't follow this convention. I think we should go through the docs an adjust all links to use the same style. Do you want to include this change in this PR? Or would you like me to take a look at this, after this is merged?

mipidsi/src/lib.rs
```diff
@@ -277,7 +277,7 @@ where
     /// Setting the vertical scroll offset shifts the vertical scroll region
     /// upwards by `offset` pixels.
     ///
-    /// Use [`set_vertical_scroll_region`] to setup the scroll region, before
+    /// Use [Self::set_vertical_scroll_region] to setup the scroll region, before
```
rfuest, line 280:
```suggestion
    /// Use [`set_vertical_scroll_region`](Self::set_vertical_scroll_region) to setup the scroll region, before
```

rfuest (changes requested):
A small stylistic nit, but it looks good otherwise. Having the docs checked in CI should improve the quality of our docs.

PR #116, opened 2024-02-14
remove `with_` prefix from Builder, fixes #113

### LINKED ISSUE #113: Removing the `with_` prefix from builder methods
Should we remove the `with_` prefix from the builder methods? The API guidelines state that the `with_` prefix should be used for constructors (https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) and builders usually don't use and prefix for the methods (https://rust-lang.github.io/api-guidelines/type-safety.html?highlight=builder#builders-enable-construction-of-complex-values-c-builder).

The naming is confusing in the latest release where [`Builder::with_model`](https://docs.rs/mipidsi/latest/mipidsi/struct.Builder.html#method.with_model) is a constructor, while other methods, like [`Builder::with_orientation`](https://docs.rs/mipidsi/latest/mipidsi/struct.Builder.html#method.with_orientation), are used to set a parameters. In the master branch `with_model` was renamed into `new`, which makes it harder to mix up the two types of methods, but I would still prefer to remove the prefix.

almindor:
Removes the `with_` prefix from `Builder`

### DIFF
--- a/mipidsi/CHANGELOG.md
+++ b/mipidsi/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- rename all `Builder::with_*` methods to versions without the `with_` prefix
+- rename all `Builder::with_*` methods to versions without the `with_` prefix to conform to the Rust API guidelines. See [this issue](https://github.com/almindor/mipidsi/issues/113) for more info
 - `options` and `error` submodule types are no longer re-exported from the main library
 - DCS commands param fields are now all consistently private with added constructors for all such commands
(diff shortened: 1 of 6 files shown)


mipidsi/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- rename all `Builder::with_*` methods to versions without the `with_` prefix
```
rfuest, line 21: Perhaps we should add the reason for this change to the changelog and migration guide, which would be to conform with the Rust API guidelines. Otherwise the users might think that we just like to break their code :wink:

almindor: done

Issue #99, opened 2024-01-23
Fix examples before next release


rfuest:
We need to remember to fix the examples before the next release. They will already have been broken by the removal of the re-exports, but I would suggest to wait until we have decided which breaking PRs we want to land before the next release.

almindor:
I believe this is the last blocker for v0.8 release.

rfuest:
The migration guide also needs to be finished before the release, it is at least missing the unified defaults for color inversion and the bumped MSRV. And we must also find a solution for #76 to fix the troubleshooting link in the new version of the docs.

I'll take some time to look at #81 in the next few days. I think that the most important changes from that PR have been merged in other smaller PRs already. But it's been so long since I've looked at that code that I can't remember the details. If I find anything useful I'll open a new PR for that change and close #81.

almindor:
I believe this was done

PR #117, opened 2024-02-19
fix spi st7789 example for v0.8

almindor:
This is the only one I know how to fix, the others hal packages don't seem to be at eh1 yet.

Related to #99 

NOTE: this cannot be merged until v0.8 is released

### DIFF
--- a/mipidsi/examples/spi-st7789-rpi-zero-w/Cargo.toml
+++ b/mipidsi/examples/spi-st7789-rpi-zero-w/Cargo.toml
@@ -4,9 +4,10 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-display-interface-spi = "0.4.1"
+display-interface-spi = "0.5"
 embedded-graphics = "0.8.0"
-mipidsi = "0.7.1"
-rppal = { version = "0.14.1", features = ["hal"] }
+embedded-hal = "1.0.0"
+mipidsi = "0.8.0"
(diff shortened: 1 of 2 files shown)


rfuest:
> This is the only one I know how to fix, the others hal packages don't seem to be at eh1 yet.

I don't know about the ESP32 one, but there is e-h 1 support for the pico. The feature to enable support is still called `eh1_0_alpha`, but the latest release uses the final version of e-h 1.

> NOTE: this cannot be merged until v0.8 is released

Because CI is failing? We will need to change CI anyway to make sure it uses the latest development version to build the examples. The simplest solution would be to use path dependencies in the examples. But this would make anyone how just wants to try one of the examples also use the latest development version. I'm not sure if that's a problem, but all solutions I can come up with have some drawback.

rursprung:
> I don't know about the ESP32 one

you can use the `eh1` feature on the latest version of the HAL

> Because CI is failing? We will need to change CI anyway to make sure it uses the latest development version to build the examples. The simplest solution would be to use path dependencies in the examples. But this would make anyone how just wants to try one of the examples also use the latest development version. I'm not sure if that's a problem, but all solutions I can come up with have some drawback.

IMHO the path dependency is the best way for such examples - it ensures that you examples always match what's currently in the repo (and anyone looking at the examples is looking at the repo). it's really confusing if the two don't match.

almindor:
superseded

PR #125, opened 2024-04-13
added ST7796 model

liebman:
Added a model for ST7796.

Previously I used the ST7789 model and set the framebuffer size via options, since thats not possible now, I created a model for it.  Its exactly ST7789 but with the framebuffer size changed.

### DIFF
--- a/mipidsi/README.md
+++ b/mipidsi/README.md
@@ -41,6 +41,7 @@ Variants that require different screen sizes and window addressing offsets are n
 * ILI9486
 * ST7735
 * ST7789
+* ST7796
 
 ## Migration
 
--- a/mipidsi/src/models/st7796.rs
+++ b/mipidsi/src/models/st7796.rs
@@ -1,0 +1,78 @@
+use display_interface::{DataFormat, WriteOnlyDataCommand};
+use embedded_graphics_core::{pixelcolor::Rgb565, prelude::IntoStorage};
+use embedded_hal::{delay::DelayNs, digital::OutputPin};
+
+use crate::{
+    dcs::{
+        BitsPerPixel, Dcs, EnterNormalMode, ExitSleepMode, PixelFormat, SetAddressMode,
+        SetDisplayOn, SetInvertMode, SetPixelFormat, SoftReset, WriteMemoryStart,
+    },
+    error::{Error, InitError},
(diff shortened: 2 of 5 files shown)


mipidsi/src/models/st7796.rs
```diff
@@ -0,0 +1,78 @@
...
+        dcs.write_command(SetPixelFormat::new(pf))?;
+        delay.delay_us(10_000);
+        dcs.write_command(EnterNormalMode)?;
+        delay.delay_us(10_000);
+        dcs.write_command(SetDisplayOn)?;
+
+        // DISPON requires some time otherwise we risk SPI data issues
+        delay.delay_us(120_000);
+
+        Ok(madctl)
```
rfuest, line 62:
```suggestion
        ST7789.init(dcs, delay, options, rst)
```

mipidsi/src/models/st7796.rs
```diff
@@ -0,0 +1,78 @@
...
+        DI: WriteOnlyDataCommand,
+        I: IntoIterator<Item = Self::ColorFormat>,
+    {
+        dcs.write_command(WriteMemoryStart)?;
+
+        let mut iter = colors.into_iter().map(Rgb565::into_storage);
+
+        let buf = DataFormat::U16BEIter(&mut iter);
+        dcs.di.send_data(buf)?;
+        Ok(())
```
rfuest, line 76:
```suggestion
        ST7789.write_pixels(dcs, colors)
```

mipidsi/README.md
```diff
@@ -41,6 +41,7 @@ Variants that require different screen sizes and window addressing offsets are n
 * ILI9486
 * ST7735
 * ST7789
+* ST7796
```
rfuest, line 44: There is another copy of this list in `lib.rs`, which should also be updated.

rfuest (changes requested):
If this model uses the same code as the ST7789 we could just reuse that implementation instead of duplicating the code.

You should also update the changelog to include the added model.

liebman:
Updated with your suggested changes.  Also added other missing models from `lib.rs`.

rfuest (approved):
LGTM and good catch with the out of sync list in `lib.rs`.

PR #127, opened 2024-04-13
Make ModelOptions fields public

### LINKED ISSUE #126: readme claims
The README claims that:

> External crates can be used to provide additional models 

However some fields in `ModelOptions` have no public accessors making that not possible, one example: `invert_colors`.

rfuest:
This PR makes the fields in `ModelOptions` public. It isn't possible to access the model options mutably from user code or external `Model` impls, which makes setters/getters unnecessary. I've also added a test to check that it is possible to implement external models and ensure that we don't remove this possibility accidentally in the future.

Fixes #126.


almindor (approved):
Thank you!

PR #130, opened 2024-04-25
Use path dependencies for examples

rfuest:
This PR changes the dependencies in the examples to use a path dependency to make sure the examples stay compatible with the latest development version of `mipidsi`. I did not test any of the changes on real hardware yet, but at they should at least all compile.

### DIFF
--- a/mipidsi/examples/spi-ili9486-esp32-c3/.cargo/config.toml
+++ b/mipidsi/examples/spi-ili9486-esp32-c3/.cargo/config.toml
@@ -7,22 +7,6 @@ rustflags = [
   # Required to obtain backtraces (e.g. when using the "esp-backtrace" crate.)
   # NOTE: May negatively impact performance of produced code
   "-C", "force-frame-pointers",
-
-  # comment the cfgs below if you do _not_ wish to emulate atomics.
-  # enable the atomic codegen option for RISCV
-  "-C", "target-feature=+a",
-  # tell the core library have atomics even though it's not specified in the target definition
-  "--cfg", "target_has_atomic_load_store",
-  "--cfg", 'target_has_atomic_load_store="8"',
-  "--cfg", 'target_has_atomic_load_store="16"',
-  "--cfg", 'target_has_atomic_load_store="32"',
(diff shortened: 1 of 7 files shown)


mipidsi/examples/spi-ili9486-esp32-c3/.cargo/config.toml
```diff
@@ -7,22 +7,6 @@ rustflags = [
...
-  "--cfg", 'target_has_atomic_load_store="8"',
-  "--cfg", 'target_has_atomic_load_store="16"',
-  "--cfg", 'target_has_atomic_load_store="32"',
-  "--cfg", 'target_has_atomic_load_store="ptr"',
-  # enable cas
-  "--cfg", "target_has_atomic",
-  "--cfg", 'target_has_atomic="8"',
-  "--cfg", 'target_has_atomic="16"',
-  "--cfg", 'target_has_atomic="32"',
-  "--cfg", 'target_has_atomic="ptr"',
```
rfuest, line 25: I'm not sure about these compiler options, but the example didn't compile with them. The latest template from the esp-rs project doesn't use these either so I guess they aren't needed, but I haven't looked into it any further.

mipidsi/examples/spi-st7789-rpi-zero-w/src/main.rs
```diff
@@ -66,15 +67,13 @@ fn main() -> ExitCode {
...
-        // width and height are switched on purpose because of the orientation
-        .with_display_size(H as u16, W as u16)
-        // this orientation applies for the Display HAT Mini by Pimoroni
-        .with_orientation(mipidsi::Orientation::LandscapeInverted(true))
-        .with_invert_colors(mipidsi::ColorInversion::Inverted)
-        .init(&mut delay, None::<OutputPin>)
+    let mut display = Builder::new(ST7789, di)
+        .display_size(W as u16, H as u16)
+        .invert_colors(ColorInversion::Inverted)
+        .init(&mut delay)
```
rfuest, line 76: The orientation is probably not the same as before, but I don't own the display hat to test it.

PR #133, opened 2024-04-26
update reset pin docs fixes #132

### LINKED ISSUE #132: Find a better place for the reset pin warning in the `Builder::init` docs
The docs for the `init` method still contain this warning about the correct state of the reset pin: https://docs.rs/mipidsi/latest/mipidsi/struct.Builder.html#warning. This place doesn't make sense anymore, now that the `init` method no longer takes the reset pin as a parameter.

Adding the warning to `Builder::reset_pin` wouldn't be very discoverable in my opinion. Perhaps the troubleshooting doc would be a better place to mention this?

almindor:


docs/TROUBLESHOOTING.md
```diff
@@ -2,6 +2,10 @@
 
 ## Display stays black/blank
 
+### Reset pin
+
+If using the reset pin it needs to be in *high* state in order for the display to operate. If it wasn't provided to the builder the user needs to ensure this is the case.
```
rfuest, line 7: I tried to clarify this a bit more and added that it is recommended to connect the reset pin to a GPIO pin:
```suggestion
The reset pin on all supported display controllers is active low, requiring it to be driven **high** in order for the display to operate. It is recommended to connect the reset pin to a GPIO pin and let this crate control the pin by passing it to the builder via the `reset_pin` method. If this isn't possible in the target application the user must make sure that the reset pin on the display controller is kept in the high state before `init` is called.

PR #135, opened 2024-05-23
Update migration guide

rfuest:
I'm not sure this is everything that should be included in the migration guide, but I cannot think of anything else at the moment.

### DIFF
--- a/mipidsi/examples/parallel_ili9341_rp_pico/Cargo.toml
+++ b/mipidsi/examples/parallel_ili9341_rp_pico/Cargo.toml
@@ -5,7 +5,7 @@ version = "0.1.0"
 
 [dependencies]
 cortex-m = "0.7.7"
-cortex-m-rt = "0.7.4"
+cortex-m-rt = "0.7.3"
 embedded-hal = "1.0.0"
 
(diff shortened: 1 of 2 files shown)


mipidsi/examples/parallel_ili9341_rp_pico/Cargo.toml
```diff
@@ -5,7 +5,7 @@ version = "0.1.0"
 
 [dependencies]
 cortex-m = "0.7.7"
-cortex-m-rt = "0.7.4"
+cortex-m-rt = "0.7.3"
```
almindor, line 8: why the downgrade?

almindor (changes requested):
LGTM apart from the cargo cortex dependency downgrade

mipidsi/examples/parallel_ili9341_rp_pico/Cargo.toml
rfuest: `0.7.4` has been yanked from crates.io: https://crates.io/crates/cortex-m-rt/0.7.4

almindor (approved):

Issue #142, opened 2024-10-18
How to speed things up (ST7789V2 + ESP32-C6-DevKitC-1)


akx:
Heya,

I'm wondering how to speed things up for an application that will likely need full-screen updates most of the time.

I have an [ESP32-C6-DevKitC-1](https://docs.espressif.com/projects/esp-dev-kits/en/latest/esp32c6/esp32-c6-devkitc-1/user_guide.html) and a [WaveShare 280x240 1.69" LCD module](https://www.waveshare.com/wiki/1.69inch_LCD_Module), using `esp-idf-svc` (at present anyway).

My current experiment code (please excuse the mess, it's an experiment so far) is

```rust
use std::collections::VecDeque;
use std::time::Instant;
use esp_idf_svc::hal::gpio::PinDriver;
use esp_idf_svc::hal::prelude::*;
use crate::led::WS2812RMT;

use embedded_graphics::{
    prelude::*,
    primitives::Rectangle,
};
use embedded_hal::spi::MODE_3;
use esp_idf_svc::hal::delay::Ets;
use esp_idf_svc::hal::spi::config::{BitOrder, Config, Duplex};
use esp_idf_svc::hal::spi::{SpiDeviceDriver, SpiDriverConfig};
use display_interface_spi::SPIInterface;
use embedded_graphics::pixelcolor::{Rgb565, Rgb888};
use mipidsi::models::ST7789;
use mipidsi::options::{ColorInversion, Orientation, RefreshOrder, Rotation};
use rand::{Rng, SeedableRng};
use rand::rngs::SmallRng;
use crate::ticktock::Ticktock;

mod led;
mod ticktock;

fn main() -> anyhow::Result<()> {
    esp_idf_svc::sys::link_patches();
    esp_idf_svc::log::EspLogger::initialize_default();
    log::info!("Hello, world!");
    let peripherals = Peripherals::take()?;
    log::info!("Initializing LED");
    let mut led = WS2812RMT::new(peripherals.pins.gpio8, peripherals.rmt.channel0)?;
    led.set_pixel(50, 50, 0)?;
    let mut rng = SmallRng::from_entropy();
    log::info!("Initializing pins");
    let spi = peripherals.spi2;
    let mut backlight = PinDriver::output(peripherals.pins.gpio23)?;
    let rst = PinDriver::output(peripherals.pins.gpio22)?;
    let dc = PinDriver::output(peripherals.pins.gpio21)?; // DC (data/command)
[...]

PR #143, opened 2024-10-21
Add direct pixel writes from u8 buffers

almindor:
Related to #142 this change will allow direct writes of `&[u8]` slice data via the `display-interface` without needing to do pixel format conversion in cases where the user knows it is safe to do so (e.g. prepared image data).

The hope is that this can get paired up with a change in `embedded-graphics` that enables such operations on a higher level (e.g. `draw_image`)

### DIFF
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -222,6 +222,10 @@ where
     /// * `ex` - x coordinate end
     /// * `ey` - y coordinate end
     /// * `colors` - anything that can provide `IntoIterator<Item = u16>` to iterate over pixel data
+    /// <div class="warning">
+    /// The `ex`and `ey` coordinates are inclusive. For example when using a rectangle
+    /// of size `320x240`, one would use `319` an `239` as `ex` and `ey` values.
+    /// </div>
     pub fn set_pixels<T>(
         &mut self,
         sx: u16,
@@ -255,7 +259,19 @@ where
     /// * `sy` - y coordinate start
     /// * `ex` - x coordinate end
     /// * `ey` - y coordinate end
-    /// * `raw_buf` - &[u8] buffer of raw pixel data in the format expected by the display.
+    /// * `raw_buf` - `&[u8]` buffer of raw pixel data in the format expected by the display.
+    /// <div class="warning">
+    /// This method requires the <b>raw_buf</b> data to be in the correct endianness
+    /// and format expected by the display.
+    ///
+    /// The method won't <b>work with a 16bit display-interface-gpio</b>, because it
+    /// pads the each byte to a u16 instead of converting each two byte chunk
+    /// into a u16. [See here for more info](https://github.com/therealprof/display-interface/blob/8fca041b0288740678f16c1d05cce21bd3867ee5/parallel-gpio/src/lib.rs#L267)
+    /// </div>
+    /// <div class="warning">
+    /// The <b>ex</b> and <b>ey</b> coordinates are inclusive. For example when using a rectangle
+    /// of size <b>320x240</b>, one would use <b>319</b> an <b>239</b> as <b>ex</b> and <b>ey</b> values.
+    /// </div>
     pub fn set_pixels_raw_u8(
         &mut self,
         sx: u16,
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -248,8 +248,6 @@ where
     /// slice has more color values than the number of pixels in the given region.
     ///
     /// This is a low level function, which isn't intended to be used in regular user code.
-    /// Consider using the [`draw_image`](https://TODO)
-    /// function from the `embedded-graphics` crate as an alternative instead (TODO)
     ///
     /// # Arguments
     ///
--- a/mipidsi/src/lib.rs
+++ b/mipidsi/src/lib.rs
@@ -239,6 +239,39 @@ where
         Ok(())
     }
 
+    ///
+    /// Sets pixel colors in a rectangular region.
+    ///
+    /// The color values from the `raw_buf` slice will be drawn to the given region starting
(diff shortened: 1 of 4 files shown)


almindor:
> I think this is a valid workaround until we have something better available in the future. We just need to make sure to document possible issues a user could run into:
> 
> * The user is responsible for getting the endianness right
> * We should document that the ranges from `sx` to `ex` and from `sy` and `ey` are inclusive (this also applies to the existing `set_pixels`)
> * Adding bounds checking for the coordinates wouldn't be a bad idea. IIRC there has been  some confusion in the past where a user tried to use `display.set_pixels(0, 0, 320, 240, colors)` instead of the correct `display.set_pixels(0, 0, 319, 239, colors)`. Without an error this problem isn't easy to catch.
> * The method won't work with a 16bit `display-interface-gpio`, because it pads the each byte to a `u16` instead of converting each two byte chunk into a `u16`. (https://github.com/therealprof/display-interface/blob/8fca041b0288740678f16c1d05cce21bd3867ee5/parallel-gpio/src/lib.rs#L267)

Sounds good, I'll add the documentation points. Do you plan to add the `draw_image` to embedded graphics? I'm wondering if I should wait for that so it's linked properly.

rfuest:
> Do you plan to add the `draw_image` to embedded graphics? I'm wondering if I should wait for that so it's linked properly.

I'm not sure when that will be added or how exactly it will be implemented. I would suggest to leave that part out and open an issue to remind us to update the comment after this has been added to e-g.

mipidsi/src/lib.rs
```diff
@@ -239,6 +243,49 @@ where
...
+    /// and format expected by the display.
+    ///
+    /// The method won't <b>work with a 16bit display-interface-gpio</b>, because it
+    /// pads the each byte to a u16 instead of converting each two byte chunk
+    /// into a u16. [See here for more info](https://github.com/therealprof/display-interface/blob/8fca041b0288740678f16c1d05cce21bd3867ee5/parallel-gpio/src/lib.rs#L267)
+    /// </div>
+    /// <div class="warning">
+    /// The <b>ex</b> and <b>ey</b> coordinates are inclusive. For example when using a rectangle
+    /// of size <b>320x240</b>, one would use <b>319</b> an <b>239</b> as <b>ex</b> and <b>ey</b> values.
+    /// </div>
```
rfuest, line 274: I think we could make this a bit clearer by tweaking a few things:
- Consistently use "raw pixel data" instead of "color values".
- Use copy or write instead of draw.
- Merge the note about being low level with the warning about the requirements for the correct data format. I think it's better to provide a reason why this should be avoided and what the alternative is.

```suggestion
    /// Copies raw pixel data to a rectangular region of the framebuffer.
    ///
    /// This method writes the pixel data stored in the `raw_buf` slice to the
    /// specified rectangular region within the display controller's
    /// framebuffer.  If `raw_buf` contains more data than required to fill the
    /// region, writing will wrap around to the top left corner after reaching
    /// the bottom right corner.
    ///
    /// <div class="warning">
    ///    
    /// This method is intended for advanced use cases where low level access to
    /// the displays framebuffer is required for performance reasons.  The
    /// caller must ensure the raw pixel data in `raw_buf` has the correct
    /// format and endianness.
    /// 
    /// For all other use cases it is recommended to instead use
    /// [`embedded-graphics`](https://docs.rs/embedded-graphics/latest/embedded_graphics/)
    /// to draw to the display.
    ///
    /// </div>
    ///
    /// <div class="warning">
    ///
    /// The method might not work the same for all `display-interface`s and is
    /// known to not work as expected when used with a
    /// [`PGPIO16BitInterface`](https://docs.rs/display-interface-parallel-gpio/latest/display_interface_parallel_gpio/struct.PGPIO16BitInterface.html)
    /// from the `display-interface-gpio` crate.
    ///
    /// </div>
    ///
    /// # Arguments
    ///
    /// * `sx` - x coordinate start
    /// * `sy` - y coordinate start
    /// * `ex` - x coordinate end (inclusive)
    /// * `ey` - y coordinate end (inclusive)
    /// * `raw_buf` - `&[u8]` buffer of raw pixel data in the format expected by the display
    ///
    /// <div class="warning">
    ///
    /// The end values of the X and Y coordinate ranges are inclusive, and no
    /// bounds checking is performed on these values. Using out of range values
    /// (e.g., passing `320` instead of `319` for a 320 pixel wide display) will
    /// result in undefined behavior.
    /// 
    /// </div>
```

mipidsi/src/models.rs
```diff
@@ -74,4 +73,26 @@ pub trait Model {
...
+    /// No pixel color format conversion, raw data is passed on directly.
+    /// <div class="warning">
+    /// This method requires the `raw_buf` data to be in the correct endianness
+    /// and format expected by the display.
+    ///
+    /// The method <b>won't work with a 16bit display-interface-gpio</b>, because it
+    /// pads the each byte to a u16 instead of converting each two byte chunk
+    /// into a u16. [See here for more info](https://github.com/therealprof/display-interface/blob/8fca041b0288740678f16c1d05cce21bd3867ee5/parallel-gpio/src/lib.rs#L267)
+    /// </div>
+    fn write_pixels_raw_u8<DI>(&mut self, dcs: &mut Dcs<DI>, raw_buf: &[u8]) -> Result<(), Error>
```
rfuest, line 88: Do we really need this method? At the moment I don't see any benefit over implementing this in `Display` directly. If we later discover that this must have a model specific implementation it's easy to add the method back in a non breaking change.

rfuest (changes requested):
Some more doc comment bike shedding.

mipidsi/src/lib.rs
almindor: Applied with rename

rfuest (approved):
LGTM

PR #145, opened 2024-11-11
Fill solid optimizations per-color

almindor:
This adds optimized (per color type) `fill_solid` paths as part of improving on issue #142.

This improved `fill_solid` fullscreen render speed on a `ST7789` using `80Mhz` SPI interface with `esp32-c6` board running at `160Mhz` CPU speed from `~48ms` to `~18ms`.

NOTE: the buffer size is arbitrarily chosen, bigger ones can get even better results, but I wanted to avoid using too much memory. If `batch` support is disabled these paths are not used. I hope we can come up with a better story for buffers overall in the future (e.g. wrt. to `display-interface` as well as this driver).

### DIFF
--- a/docs/MIGRATION.md
+++ b/docs/MIGRATION.md
@@ -1,5 +1,15 @@
 # Migration guide for `mipidsi` crate
 
+## v0.8 -> 0.9
+
+### Users
+
+* No changes
+
+### Model writers
+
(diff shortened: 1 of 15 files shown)


almindor:
I'm also considering refactoring this using https://crates.io/crates/disjoint_impls/0.8.0 which should allow removal of the extra `Model` method and just doing it directly in the graphics module.

almindor:
Closing in favour of #146 , however if for some reason we don't want to use the specialization "hack" we can fall back to this.

PR #147, opened 2024-11-15
Make fill_params_buf infallible

GrantM11235:
The error variant was never used.


almindor (approved):
Thanks, nice simplification

PR #152, opened 2024-12-25
ili9341: Add Bgr565 support

cazou:
Some devices like some Waveshare LCD Modules use the BGR565 pixel format.

Make sure it is supported.

### DIFF
--- a/mipidsi/src/interface/mod.rs
+++ b/mipidsi/src/interface/mod.rs
@@ -1,7 +1,7 @@
 //! Interface traits and implementations
 
 mod spi;
-use embedded_graphics_core::pixelcolor::{Rgb565, Rgb666};
+use embedded_graphics_core::pixelcolor::{Bgr565, Rgb565, Rgb666};
 pub use spi::*;
 
 mod parallel;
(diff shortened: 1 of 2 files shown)


rfuest:
I believe that the ILI9341 doesn't support BGR colors in the way this PR implements them. The pixel data is expected to always be encoded as RGB when it is transmitted via SPI or a parallel interface.

A common reason for flipped R and B channels is an incorrectly configured subpixel order: https://github.com/almindor/mipidsi/blob/master/docs/TROUBLESHOOTING.md#wrong-subpixel-order. Does changing this setting fix the colors on your display module if you use `Rgb565`?

cazou:
> I believe that the ILI9341 doesn't support BGR colors in the way this PR implements them. The pixel data is expected to always be encoded as RGB when it is transmitted via SPI or a parallel interface.
> 
> A common reason for flipped R and B channels is an incorrectly configured subpixel order: https://github.com/almindor/mipidsi/blob/master/docs/TROUBLESHOOTING.md#wrong-subpixel-order. Does changing this setting fix the colors on your display module if you use `Rgb565`?

Ha indeed, I should have read that part :p 
Sorry for the noise

PR #154, opened 2025-01-04
Fix: Rgb666 Byte Conversion

### LINKED ISSUE #153: Rgb666 Color Bytes Seem Incorrect
### Problem
I was testing the new changes on `master` and found that the new interface PR (#149) may have broken Rgb666 colors. The last commit with the correct colors was `e808fb9`. Using Rgb565 still works correctly.

I suspect that the problem might be in the new [`rgb666_to_bytes` function](https://github.com/almindor/mipidsi/blob/2681465f4893458be401ac7648d9c44a50aa428f/mipidsi/src/interface/mod.rs#L75). I found the [old version](https://github.com/almindor/mipidsi/blob/e808fb91758af54fc746d09c092a52c8cb65a25d/mipidsi/src/models/ili934x.rs#L72) and decided to compare their results for `Rgb666::RED` and found they were different:
```rs
// New
[12, 192, 0]
// Old
[252, 0, 0]
```
I'm unsure if this is the issue or not.

I'm using an `esp32-wroom-32e` over SPI w/`ili9341`. I tested with the test image and no combination of color ordering or inversion resolved the problem. On my previously correct ordering (Bgr) & inversion (Normal), red was light green, green was dark green, and blue was blue.

DogeDark:
Fixes #153 

~~Ideally, I'd like to reuse the original `PixelColor` trait, but for some reason, it's not registering the import.~~ Got it working!

### DIFF
--- a/mipidsi/src/interface/mod.rs
+++ b/mipidsi/src/interface/mod.rs
@@ -1,7 +1,7 @@
 //! Interface traits and implementations
 
 mod spi;
-use embedded_graphics_core::pixelcolor::{Rgb565, Rgb666};
+use embedded_graphics_core::pixelcolor::{Rgb565, Rgb666, RgbColor};
 pub use spi::*;
 
 mod parallel;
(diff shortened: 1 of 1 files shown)


rfuest (approved):
LGTM

almindor (approved):
Thank you!

PR #157, opened 2025-01-11
Add RM67162 display driver implementation

psytraxx:
**Specs**:
https://www.orientdisplay.com/wp-content/uploads/2022/08/RM67162.pdf
**Used for example in** 
https://github.com/Xinyuan-LilyGO/LilyGo-AMOLED-Series

**Usage example:**
Still with the driver in my own crate - But since i use mipidsi crate in all my projects anyway i thought i might try to publish the driver through this crate 
https://github.com/psytraxx/s3-display-amoled-touch-rs


rfuest:
Somehow the display format doesn't match the datasheet. It's a bit odd that the datasheet claims that the framebuffer is 480x480 pixels, but also mentions one supported display resolution that is 540 pixel high (greater than the max. framebuffer dimension).

Did you verify that all display orientations work correctly when you draw the builtin `TestImage`?

psytraxx:
Thanks for the Test Image hint.  I actually swapped width and height which first led to distorted test image. That is fixed now and this is the output after my latest change .

image
image
image
image

This is what took as reference for the init sequence and dimensions:
https://github.com/Xinyuan-LilyGO/LilyGo-AMOLED-Series/blob/master/src/initSequence.h#L29C16-L29C22

I am not an expert but from I understand this display  documentation 

Display resolution option 
 480RGB x 480 with 480x24-bits x 480 GRAM 
 400RGB x 400 with 400x24-bits x 400 GRAM 
 360RGB x 480 with 360x24-bits x 480 GRAM 
 320RGB x 320 with 320x24-bits x 320 GRAM 
 320RGB x 480 with 320x24-bits x 480 GRAM 
 272RGB x 480 with 272x24-bits x 480 GRAM 
 240RGB x 240 with 240x24-bits x 240 GRAM 
 240RGB x 320 with 240x24-bits x 320 GRAM 
 180RGB x 360 with 180x24-bits x 360 GRAM 
 180RGB x 540 with 180x24-bits x 540 GRAM 
 128RGB x 432 with 128x24-bits x 432 GRAM 
 Display data RAM (frame memory): 480 x480 x 24-bits = 5,529,600 bits 

the actually resolution is not listed but as long as the RAM fits there shouldn't be a problem right?

mipidsi/src/models/rm67162.rs
```diff
@@ -0,0 +1,68 @@
...
+        let madctl = SetAddressMode::from(options);
+
+        di.write_raw(0xFE, &[0x04])?; // SET APGE3
+        di.write_raw(0x6A, &[0x00])?;
+        di.write_raw(0xFE, &[0x05])?; // SET APGE4
+        di.write_raw(0xFE, &[0x07])?; // SET APGE6
+        di.write_raw(0x07, &[0x4F])?;
+        di.write_raw(0xFE, &[0x01])?; // SET APGE0
+        di.write_raw(0x2A, &[0x02])?; // Set column start address
+        di.write_raw(0x2B, &[0x73])?; // Set row start address
```
rfuest, line 44: I don't think these comments are accurate. At this point in the init sequence the controller is in the Manufacture Command Set Page 0, which probably means that `0x2A` and `0x2B` are different commands. Both commands would otherwise require 4 bytes as parameters.

My suggestion would be to just remove the comments, because we have no idea what these do:
```suggestion
        di.write_raw(0x2A, &[0x02])?;
        di.write_raw(0x2B, &[0x73])?;
```

mipidsi/src/models/rm67162.rs
```diff
@@ -0,0 +1,68 @@
...
+        di.write_raw(0xFE, &[0x0A])?; // SET APGE9
+        di.write_command(SetDisplayOn)?;
```
rfuest, line 46: This is another case where the controller is in another command page. The original init sequence in the repo you have linked does also send a parameter (`0x10`) for this command: https://github.com/Xinyuan-LilyGO/LilyGo-AMOLED-Series/blob/8c72b786373fbaef46ce35a6db924d6e16a0c3ec/src/initSequence.cpp#L539

rfuest (changes requested):
> the actually resolution is not listed but as long as the RAM fits there shouldn't be a problem right?

There might be a problem if someone tries to use this code for another display size. My guess is that somewhere in the undocumented manufacturer commands the framebuffer gets reshaped to allow for the display resolution of your display.

My suggestion would be to document that this currently has only been tested with this resolution and might fail if used with another display size. We can revisit the code later, when someone reports issues with using it on another display.

Could you also add an entry to the changelog?

almindor (changes requested):
Thank you for providing this driver. Please add CHANGELOG and README entries

mipidsi/src/models/rm67162.rs
psytraxx: I removed the comments - added the missing command and the comment about the resolution support.

psytraxx:
Thanks for your review.  I pushed my changes which should adress all the issues raised. I also tested a bit more with the device and i still looks good. That displayon command which sends no value in your version might be a bug in the orginal code - i checked the documentation and there it says that there are no arguments - which is probably the reason why the replacement with displayOn did work 
image

rfuest:
> Thanks for your review. I pushed my changes which should adress all the issues raised. I also tested a bit more with the device and i still looks good. That displayon command which sends no value in your version might be a bug in the orginal code - i checked the documentation and there it says that there are no arguments - which is probably the reason why the replacement with displayOn did work

I'm not so sure about that. I've found some other code which also sends the `0x2A` command with a parameter, while the controller is in the manufacturer command code: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/display/display_rm67162.c#L50 If both versions work I would prefer to use the raw version with the parameter.

mipidsi/src/models/rm67162.rs
```diff
@@ -0,0 +1,76 @@
...
+        SetInvertMode, SetPixelFormat,
+    },
+    interface::Interface,
+    options::ModelOptions,
+};
+
+use super::Model;
+
+/// RM67162 AMOLED display driver implementation
+/// Supports:
```
rfuest, line 16: There needs to be a new line between the first line and the rest of the text. Otherwise the struct list in the docs is broken:

image


```suggestion
/// RM67162 AMOLED display driver implementation
///
/// Supports:
```

rfuest (changes requested):
The docs that describe the limitations to a specific display model are great, but there is one small issue with the docs.

psytraxx:
> I'm not so sure about that. I've found some other code which also sends the `0x2A` command with a parameter, while the controller is in the manufacturer command code: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/display/display_rm67162.c#L50 If both versions work I would prefer to use the raw version with the parameter.

wasn't this inital comment of yours about the displayon command which has a an argument in the 
initial arduino  init sequence which but existing displayon hasn't this ?

anyways - its reverted to  
di.write_raw(0x29, &[0x10])?;

both work - i tested this .

rfuest:
> wasn't this inital comment of yours about the displayon command which has a an argument in the
> initial arduino init sequence which but existing displayon hasn't this ?

My initial comment was that the first time the `0x2A` command is sent it isn't really a `DisplayOn` command, but something else from the undocumented manufacturer command set. That variant seems to require a parameter, in contrast to the `DisplayOn` command. But without any proper documentation by the manufacturer of the IC this is only an educated guess, which is based on the other init code implementations for the same IC. In the Zephyr code I linked above the `0x2A` command is even sent multiple times, but each time with another active command set and another parameter value.

rfuest (approved):

almindor (approved):

PR #159, opened 2025-01-14
Remove nb and rename interface module file

almindor:
Just some minor housekeeping. I just noticed we don't use the `nb` crate at all and wanted to also unify the new interface module file name to follow project naming conventions (I dislike mod.rs as it becomes hard to know which one it is in tabs etc.)


rfuest (approved):
I have two suggestions for the changelog, but they are a bit subjective and I'm OK with merging it anyway if you prefer the current version.

PR #160, opened 2025-01-14
Support BGR565 Variant of the ILI9341 display

Fl0h0:
My ILI9341 (Board rev v1.2) uses the BGR color format which this change would now support.

### DIFF
--- a/mipidsi/src/interface/mod.rs
+++ b/mipidsi/src/interface/mod.rs
@@ -1,7 +1,7 @@
 //! Interface traits and implementations
 
 mod spi;
-use embedded_graphics_core::pixelcolor::{Rgb565, Rgb666, RgbColor};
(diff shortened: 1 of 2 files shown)


rfuest:
Thanks for the PR, but this change isn't necessary and the same change was previously submitted a few weeks ago as https://github.com/almindor/mipidsi/pull/152. See [my comment](https://github.com/almindor/mipidsi/pull/152#issuecomment-2562068804) on that PR for additional info. Let us know if changing the setting for the color order also works in your use case.

Fl0h0:
Oh sorry I totally missed that and did not look through the closed issues.

rfuest:
> Oh sorry I totally missed that and did not look through the closed issues.

No problem. This has actually been helpful, because it has confirmed my suspicion that our troubleshooting guide is easy to miss (#161).

Issue #158, opened 2025-01-12
Also supports ILI9488


NotABadger:
No problem here, Just wanted to mention that I've tested this on an ILI9488 with the mipidsi::models::ILI9486Rgb666 instance using the Generic8BitBus. and BRG color order.
Thanks you for contributing to open source!

rfuest:
Thanks for letting us know.

@almindor: Should we add a ILI9488 type to make this an officially supported model? It's a fairly common IC and at the moment it wouldn't be much of a maintenance burden. Having a separate type would allow us to add a `Rgb888` mode in the future, which isn't available in the ILI9486.

NotABadger:
Today I had a morning to hobby around, I've also tried the SPI interface, which works.
As a (relatively) young embedded engineer not experienced in display busses, I wanted to "drag race"  the different interfaces, so I wrote a test program, executed it on the MIPI type DBI 8-bit (as mentioned above) which took up to 7.1 seconds.
After that I tried to do the same via the SPI interface, but I didn't known which clock frequency to use so I then read the datasheets of the ILI9488, and ILI9486.  Which also don't mention a frequency or speed. The ESP32-S3 can run up to 80 MHz according to the Datasheet. After some experimenting I managed to run it up to 60 MHz on a breadboard with the ILI9488, and complete the same test program in 1.42 sec (and I accidentally drag raced RGB565 vs RGB666 on SPI, because I didn't read the SPI part of the datasheet beforehand).

I was wondering, if I could toy with the clock frequency of the generic bus and then I read the ili9486.rs file, and found that 120_000 us (or 120 ms) is used between write cycles.  May I know where that value came from? I'm not criticizing or anything, just curious about these type of interfaces and trying to learn. Feel free to just comment with links/articles/datasheets. 

Below is my drag race code in case you are interested, this is running on an ESP32-S3 using the esp-idf framework.


```
fn main() {
    // It is necessary to call this function once. Otherwise some patches to the runtime
    esp_idf_svc::sys::link_patches();
    // Bind the log crate to the ESP Logging facilities
    esp_idf_svc::log::EspLogger::initialize_default();

    log::info!("Display test");

    //Setting up:
    // Data/Command (D/C) output pin, Reset Pin.
    let periph = Peripherals::take().unwrap();
    let lcd_dc = PinDriver::output(periph.pins.gpio19).unwrap();
    let lcd_reset = PinDriver::output(periph.pins.gpio40).unwrap();
[...]

PR #166, opened 2025-01-18
Add support for ILI9488

rfuest:
This PR adds official support for the ILI9488. As reported in #158 it did already work by using the ILI9486 code.


rfuest:
I believe the warnings about RGB565 not being supported via SPI aren't correct. This might only apply to 3 line serial and not 4 line serial. I'll need to check this again before the PR can be merged.

NotABadger:
Using the [ILI9488 datasheet ](https://www.alldatasheet.com/datasheet-pdf/pdf/1131840/ETC1/ILI9488.html), on page 119 (3 wire SPI) and page 121 (4 wire spi) it mentions that it supports 3 bit input or 18 bit input. That's why I mentioned it.

almindor (approved):
LGTM

rfuest:
I've removed the warning about the color format, because this will be handled by the solution to issue #167.

Issue #167, opened 2025-01-18
Add physical connection type to Interface trait


rfuest:
Some display controllers support different color depth with different connection types. The ILI9488, for example, supports different color depth for SPI and parallel connections.

We've never been able to detect if the combination between physical connection and chosen color depth is valid. But now that we no longer depend on `display-interface` we could add additional meta information to the `Interface` trait to detect these kinds of configuration errors and alert the user about them.

almindor:
How do you propose to model this? A const parameter with the `Model` type-bound to a "typeflag" of implementors of the Interface trait?

rfuest:
I hadn't really thought about the details of the implementation when I opened the issue. But I think adding a const to the `Interface` trait would be a good idea like:
```rust
trait Interface {
   const KIND: InterfaceKind;
   ...
}

#[non_exhaustive]
enum InterfaceKind {
   Spi,
   Parallel8080,
}
```

The check for valid configuration would then be performed by the `Model`s `init` method. This is more flexible than trying to find a type level solution for this problem. The `init` method would need to be changed to return an `InitError` and a `InvalidConfiguration` variant added to that enum.

rfuest:
This has been implemented in #169.

PR #169, opened 2025-01-22
Add KIND contant to Interface trait

rfuest:
This PR adds the interface kind constant to the `Interface` trait as discussed in #167.

I've decided to not extend the `InitError` enum and just panic instead. It should only be possible to encounter this error during development time, when the programmer wasn't aware that the combination of controller model, color format, and interface kind isn't supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - added `RM67162` model support
 - made `InitError` visible
 - added `ILI9488` model support
+- added `KIND` constant to `Interface` trait to detect invalid model, color format, and interface combinations
 
 ## Removed
 
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -19,6 +19,9 @@ pub trait Interface {
     /// Error type
     type Error: core::fmt::Debug;
 
+    /// Kind
+    const KIND: InterfaceKind;
+
     /// Send a command with optional parameters
     fn send_command(&mut self, command: u8, args: &[u8]) -> Result<(), Self::Error>;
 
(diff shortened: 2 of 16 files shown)


almindor (approved):
LGTM!

almindor (changes requested):
Hmm actually thinking about this, I wonder if forcing panic is the wrong call here. I think I'd sleep better if we `Model::init` errored out properly in this case.

I'm not sure how to best approach that given how `Interface::Error` is currently generic. My main concern is that kernels can have basic expectations such as "drivers do not panic" and I think that should be upheld, even in init code.

rfuest:
> Hmm actually thinking about this, I wonder if forcing panic is the wrong call here. I think I'd sleep better if we `Model::init` errored out properly in this case.
> 
> I'm not sure how to best approach that given how `Interface::Error` is currently generic. My main concern is that kernels can have basic expectations such as "drivers do not panic" and I think that should be upheld, even in init code.

I don't think that this would have been an issue, because the compiler would have been able to resolve the assertions at compile time and remove them from the final binary.

I've changed the code to return a `ModelInitError` from `Model::init`, which is a variant of `InitError` that doesn't contain the reset pin error. Using `InitError` directly would have required adding an additional type parameter for the reset pin error to `Model::init`. This didn't feel right because `Model::init` has no access to the reset pin.

If you are concerned about possible panics we should also look at `Builder::display_size`, which is much more likely to panic IMO, because the parameters might not be known at compile time. We could move the size check to `Builder::init` and also return a `InitError::InvalidConfiguration` in that case. But this is outside the scope of this PR.

almindor (approved):

almindor:
> If you are concerned about possible panics we should also look at `Builder::display_size`, which is much more likely to panic IMO, because the parameters might not be known at compile time. We could move the size check to `Builder::init` and also return a `InitError::InvalidConfiguration` in that case. But this is outside the scope of this PR.

Agreed, I'll make an issue for that

PR #172, opened 2025-01-27
remove panics and fix tests to reflect

almindor:
Removes panics from `Builder` and refactors into `InitError`

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -108,9 +108,7 @@ where
 
     /// Sets the display size.
     ///
-    /// # Panics
     ///
-    /// Panics if `width` or `height` is 0.
     #[must_use]
     pub fn display_size(mut self, width: u16, height: u16) -> Self {
         self.options.display_size = (width, height);
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -113,8 +113,6 @@ where
     /// Panics if `width` or `height` is 0.
     #[must_use]
     pub fn display_size(mut self, width: u16, height: u16) -> Self {
-        assert!(width != 0 && height != 0);
-
         self.options.display_size = (width, height);
         self
     }
@@ -162,8 +160,20 @@ where
         let (width, height) = to_u32(self.options.display_size);
         let (offset_x, offset_y) = to_u32(self.options.display_offset);
         let (max_width, max_height) = to_u32(MODEL::FRAMEBUFFER_SIZE);
-        assert!(width + offset_x <= max_width);
-        assert!(height + offset_y <= max_height);
+
+        if width == 0 || height == 0 {
+            return Err(InitError::InvalidConfiguration(
+                "invalid display dimensions",
+            ));
+        }
+
+        if width + offset_x > max_width {
+            return Err(InitError::InvalidConfiguration("display too wide"));
+        }
+
+        if height + offset_y > max_height {
+            return Err(InitError::InvalidConfiguration("display too tall"));
+        }
 
         match self.rst {
             Some(ref mut rst) => {
@@ -206,14 +216,19 @@ pub enum InitError<DI, P> {
     /// This error is returned when the configuration passed to the builder is
     /// invalid. For example, when the combination of bit depth and interface
     /// kind isn't supported by the selected model.
-    InvalidConfiguration,
+    InvalidConfiguration(&'static str),
 }
+
+// unify msg for tests
+pub(crate) const MODEL_INTERFACE_MISMATCH_ERROR_MSG: &str = "display model and interface mismatch";
 
 impl<DiError, P> From<ModelInitError<DiError>> for InitError<DiError, P> {
     fn from(value: ModelInitError<DiError>) -> Self {
         match value {
             ModelInitError::Interface(e) => Self::Interface(e),
-            ModelInitError::InvalidConfiguration => Self::InvalidConfiguration,
+            ModelInitError::InvalidConfiguration => {
+                Self::InvalidConfiguration(MODEL_INTERFACE_MISMATCH_ERROR_MSG)
+            }
         }
     }
 }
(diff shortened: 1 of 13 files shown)


src/models.rs
```diff
@@ -90,7 +90,7 @@ mod tests {
     impl Model for OnlyOneKindModel {
         type ColorFormat = Rgb565;
 
-        const FRAMEBUFFER_SIZE: (u16, u16) = (0, 0);
+        const FRAMEBUFFER_SIZE: (u16, u16) = (16, 16);
```
almindor, line 93: This was done since it actually fails with 0, 0 now (as it should).

NOTE: It'd be great if we cound enforce > 0 on this constant but I guess it's not compile time possible?

rfuest (changes requested):
I think the change from panicing to errors should also be mentioned in the changelog.

src/builder.rs
```diff
@@ -206,14 +216,19 @@ pub enum InitError<DI, P> {
     /// This error is returned when the configuration passed to the builder is
     /// invalid. For example, when the combination of bit depth and interface
     /// kind isn't supported by the selected model.
-    InvalidConfiguration,
+    InvalidConfiguration(&'static str),
```
rfuest, line 219: I did think about this some more and I came to the conclusion that an enum would work better than a string. By using an enum we wouldn't have to store the error strings in flash and we could instead add doc comments to the enum variants to provide better information to the user. By making the enum `non_exhaustive` we would still be flexible and could always add new error kinds without causing a breaking change.

We might still want to have a string variant to allow external `Model` impls to report other kinds of errors.

Something like this:
```rust
pub enum InitError<...> {
   ...,
   InvalidConfiguration(pub ConfigurationError),
}

#[non_exhaustive]
pub enum ConfigurationError {
   InvalidDisplaySize,
   InvalidColorFormat,
   Other(&'static str),
}
```

almindor: refactored, good point about taking the string memory bytes.

src/builder.rs
```diff
@@ -206,14 +217,30 @@ pub enum InitError<DI, P> {
...
+#[derive(Debug)]
+pub enum ConfigurationError {
+    /// Display Model and Interface kind are mismatching
+    ModelInterfaceMismatch,
+    /// Display dimensions provided in [Builder::display_size] were invalid, e.g. width or height of 0
+    DisplayDimensions,
+    /// Display width from [Builder::display_size] + offset_x from [Builder::display_offset] > maximum width from framebuffer definition
+    XOffsetOverflow,
+    /// Display height from [Builder::display_size] + offset_y from [Builder::display_offset] > maximum height from framebuffer definition
+    YOffsetOverflow,
```
rfuest, line 234: This is probably a bit nitpicky, but I don't like the name `XOffsetOverflow, because in my mind it implies that the offset is overflowing.

How about this? I don't think we need to be as fine grained with our error reporting and instead offer a bit more explanation in the doc comment.

```suggestion
    /// Invalid display offset.
    ///
    /// The active display area, defined by
    /// [`display_size`](Builder::display_size) and
    /// [`display_offset`](Builder::display_offset), extends beyond the boundaries of
    /// the controller's framebuffer. To resolve this, reduce the offset to a
    /// maximum value of [`FRAMEBUFFER_SIZE`](Model::FRAMEBUFFER_SIZE) minus
    /// [`display_size`](Builder::display_size).
    InvalidOffset,
```

src/builder.rs
```diff
@@ -206,14 +217,30 @@ pub enum InitError<DI, P> {
...
+}
+
+/// Specifics of [InitError::InvalidConfiguration] if configuration was found invalid
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum ConfigurationError {
+    /// Display Model and Interface kind are mismatching
+    ModelInterfaceMismatch,
+    /// Display dimensions provided in [Builder::display_size] were invalid, e.g. width or height of 0
+    DisplayDimensions,
```
rfuest, line 230: I think we should keep the naming consistent and use "display size" and not "display dimensions".

```suggestion
    /// Invalid display size.
    /// 
    /// At least one dimension of the [`display_size`](Builder::display_size) is
    /// invalid. The display size must be larger than zero and less or equal to
    /// the [`FRAMEBUFFER_SIZE`](Model::FRAMEBUFFER_SIZE) for the selected
    /// model.
    InvalidSize,   
```

src/builder.rs
```diff
@@ -206,14 +217,30 @@ pub enum InitError<DI, P> {
...
-    InvalidConfiguration,
+    InvalidConfiguration(ConfigurationError),
+}
+
+/// Specifics of [InitError::InvalidConfiguration] if configuration was found invalid
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum ConfigurationError {
+    /// Display Model and Interface kind are mismatching
+    ModelInterfaceMismatch,
```
rfuest, line 228: I think it's important to mention that changing the color format might fix this issue:
```suggestion
    /// Unsupported interface kind.
    /// 
    /// The chosen interface isn't supported by the selected model. Note that
    /// some controller models don't support all combinations of physical
    /// interface and color formats. To resolve this, try to use another color
    /// format if available (e.g.
    /// [`ILI9486Rgb666`](crate::models::ILI9486Rgb666) instead of
    /// [`ILI9486Rgb565`](crate::models::ILI9486Rgb565) if you use a SPI
    /// connection).
    UnsupportedInterface,
```

rfuest (changes requested):

src/builder.rs
```diff
@@ -206,14 +217,30 @@ pub enum InitError<DI, P> {
...
 }
 
 impl<DiError, P> From<ModelInitError<DiError>> for InitError<DiError, P> {
     fn from(value: ModelInitError<DiError>) -> Self {
         match value {
             ModelInitError::Interface(e) => Self::Interface(e),
-            ModelInitError::InvalidConfiguration => Self::InvalidConfiguration,
+            ModelInitError::InvalidConfiguration => {
+                Self::InvalidConfiguration(ConfigurationError::ModelInterfaceMismatch)
+            }
```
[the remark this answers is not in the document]
almindor: I don't think changing `ModelInitError` would be helpful here, as `Interface(DiError)` would then be confusing with the new name.

LMK if the current version makes sense. `ModelInitError` now has `InvalidConfiguration` which gets translated to `ConfigurationError::UnsupportedInterface`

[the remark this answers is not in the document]
rfuest: > I don't think changing `ModelInitError` would be helpful here, as `Interface(DiError)` would then be confusing with the new name.

That would only be an issue if we used the second variant I suggested, which wasn't my preferred solution anyway. I don't see any reason not to change `ModelInitError::InvalidConfiguration` to `ModelInitError::InvalidConfiguration(ConfigurationError)` to make it possible to report other kinds of configuration errors from `Model::init` impls in the future.
One example where this might become useful in the future is the recently added RM67162 support. This controller doesn't have a fixed framebuffer size and it seems like the shape of the framebuffer can be adjusted via manufacturer specific commands. This is very specific to that model and `Builder::init` can't fully check if the display size in this case. But `Model::init` would be able to perform additional checks and could, for example, return `Err(ModelInitError(ConfigurationErrorr::InvalidDisplaySize))`.

src/builder.rs
rfuest: Why didn't you use this doc comment, but the other two?

src/builder.rs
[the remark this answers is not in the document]
almindor: fixed

src/builder.rs
almindor: fixed

src/builder.rs
```diff
@@ -162,8 +160,20 @@ where
         let (width, height) = to_u32(self.options.display_size);
         let (offset_x, offset_y) = to_u32(self.options.display_offset);
         let (max_width, max_height) = to_u32(MODEL::FRAMEBUFFER_SIZE);
-        assert!(width + offset_x <= max_width);
-        assert!(height + offset_y <= max_height);
+
```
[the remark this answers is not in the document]
almindor: fixed

almindor:
@rfuest I think all the mentioned issues should be addressed now

src/builder.rs
```diff
@@ -206,14 +215,39 @@ pub enum InitError<DI, P> {
...
+pub enum ConfigurationError {
+    /// Unsupported interface kind.
+    ///
+    /// The chosen interface isn't supported by the selected model. Note that
+    /// some controller models don't support all combinations of physical
+    /// interface and color formats. To resolve this, try to use another color
+    /// format if available (e.g. [`ILI9486Rgb666`](crate::models::ILI9486Rgb666) instead of
+    /// [`ILI9486Rgb565`](crate::models::ILI9486Rgb565) if you use a SPI connection)
+    UnsupportedInterface,
+    /// Invalid display dimentions
```
rfuest, line 233: There's a typo in "dimensions":
```suggestion
    /// Invalid display dimension.
```

But perhaps it would be better to call it "size" to be consistent with the name of the corresponding setting:
```suggestion
    /// Invalid display size.
```

src/builder.rs
```diff
@@ -206,14 +215,39 @@ pub enum InitError<DI, P> {
...
+    ///
+    /// Display dimensions provided in [Builder::display_size] were invalid, e.g. width or height of 0
+    InvalidDisplaySize,
+    /// Invalid display offset.
+    ///
+    /// The active display area, defined by [`display_size`](Builder::display_size) and
+    /// [`display_offset`](Builder::display_offset), extends beyond the boundaries of
+    /// the controller's framebuffer. To resolve this, reduce the offset to a maximum value of
+    /// [`FRAMEBUFFER_SIZE`](Model::FRAMEBUFFER_SIZE) minus [`display_size`](Builder::display_size).
+    InvalidOffset,
```
rfuest, line 243: I would suggest to make this consistent with `InvalidDisplaySize` and the corresponding `display_offset` property:
```suggestion
    InvalidDisplayOffset,
```

rfuest (approved):
Two small nits, but otherwise it looks good.

PR #173, opened 2025-02-03
add SpiInterfaceAsync with flush

almindor:
This PR adds a buffered `SpiInterfaceAsync` struct and `FlushingInterface` trait to enable fully buffered operations in async context with use of manual `flush()` method.

NOTE: I'm unsure if we want to generalize buffering over ALL interfaces with something like `with_framebuffer()` constructor additions since performance wise it can be an improvment if the MCU has the RAM to spare, even without async.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,6 +31,12 @@ jobs:
       - name: Run tests
         run: |
           cargo test
+      - name: Run tests with all features
+        run: |
+          cargo test --all-features
+      - name: Run tests without default features
+        run: |
+          cargo test --no-default-features
 
   # On macOS and Windows, we at least make sure that the crate builds and links.
   build-other:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- added `async` feature with `SpiInterfaceAsync` implementation behind it
 - added `SpiInterface::release`
 - added `RM67162` model support
 - made `InitError` visible
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,14 +12,20 @@ documentation = "https://docs.rs/mipidsi"
 rust-version = "1.75"
 
 [dependencies]
-embassy-futures = "0.1.1"
 embedded-graphics-core = "0.4.0"
 embedded-hal = "1.0.0"
-embedded-hal-async = "1.0.0"
 
 [dependencies.heapless]
 optional = true
 version = "0.8.0"
+
+[dependencies.embassy-futures]
+optional = true
+version = "0.1.1"
+
+[dependencies.embedded-hal-async]
+optional = true
+version = "1.0.0"
 
 [dev-dependencies]
 embedded-graphics = "0.8.1"
@@ -27,6 +33,7 @@ embedded-graphics = "0.8.1"
 [features]
 default = ["batch"]
 batch = ["heapless"]
+async = ["embassy-futures", "embedded-hal-async"]
 
 [workspace]
 members = ["mipidsi-async"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,10 @@ documentation = "https://docs.rs/mipidsi"
 rust-version = "1.75"
 
 [dependencies]
+embassy-futures = "0.1.1"
 embedded-graphics-core = "0.4.0"
 embedded-hal = "1.0.0"
+embedded-hal-async = "1.0.0"
 
 [dependencies.heapless]
 optional = true
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -1,8 +1,10 @@
 //! Interface traits and implementations
+use embedded_graphics_core::pixelcolor::{Rgb565, Rgb666, RgbColor};
 
 mod spi;
-use embedded_graphics_core::pixelcolor::{Rgb565, Rgb666, RgbColor};
+mod spi_async;
 pub use spi::*;
+pub use spi_async::*;
 
 mod parallel;
 pub use parallel::*;
@@ -41,6 +43,12 @@ pub trait Interface {
         pixel: [Self::Word; N],
         count: u32,
     ) -> Result<(), Self::Error>;
+}
+
+/// Pixel interface that writes to a intermediate buffer and provides an async flush
+pub trait FlushingInterface: Interface {
+    /// Flushes stored pixel data to the interface
+    fn flush(&mut self) -> impl core::future::Future<Output = Result<(), Self::Error>>;
 }
 
 impl<T: Interface> Interface for &mut T {
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -1,13 +1,15 @@
 //! Interface traits and implementations
 use embedded_graphics_core::pixelcolor::{Rgb565, Rgb666, RgbColor};
 
-mod spi;
-mod spi_async;
-pub use spi::*;
-pub use spi_async::*;
-
 mod parallel;
+mod spi;
+#[cfg(feature = "async")]
+mod spi_async;
+
 pub use parallel::*;
+pub use spi::*;
+#[cfg(feature = "async")]
+pub use spi_async::*;
 
 /// Command and pixel interface
 pub trait Interface {
--- a/src/interface/spi_async.rs
+++ b/src/interface/spi_async.rs
@@ -1,4 +1,6 @@
-use embedded_hal::{digital::OutputPin, spi::SpiDevice};
+use embassy_futures::block_on;
+use embedded_hal::digital::OutputPin;
+use embedded_hal_async::spi::SpiDevice;
 
 use super::{FlushingInterface, Interface, InterfaceKind, SpiError};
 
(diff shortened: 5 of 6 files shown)


J-Guevara12:
Looks great! Cannot wait to see it merged to the main branch

almindor:
Example code on esp32-c6 MCU:
```rust
#![no_std]
#![no_main]

use embassy_embedded_hal::shared_bus::asynch::spi::SpiDevice;
use embassy_executor::Spawner;
use embassy_sync::{blocking_mutex::raw::NoopRawMutex, mutex::Mutex};
use embassy_time::{Duration, Timer};
use embedded_graphics::{
    draw_target::DrawTarget,
    pixelcolor::Rgb565,
    prelude::{Point, RgbColor, *},
    primitives::{Circle, PrimitiveStyleBuilder},
    Drawable,
};
use esp_backtrace as _;
use esp_hal::{
    clock::{Clocks, CpuClock},
    delay::Delay,
    dma::{DmaRxBuf, DmaTxBuf},
    dma_buffers,
    gpio::{Level, Output},
    spi::{
        master::{Config, Spi},
        Mode,
    },
    time::{self, RateExtU32},
};
use log::info;
use mipidsi::{
    interface::SpiInterfaceAsync,
    models::ST7789,
    options::{ColorInversion, Orientation, Rotation},
    Builder, TestImage,
};

#[esp_hal_embassy::main]
async fn main(spawner: Spawner) {
    let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
    let peripherals = esp_hal::init(config);
    let mut delay = Delay::new();

    esp_println::logger::init_logger_from_env();
    info!("running at {}", Clocks::get().cpu_clock);

    let mosi = peripherals.GPIO23;
    let miso = peripherals.GPIO13;
    let sclk = peripherals.GPIO22;
    let cs_display = Output::new(peripherals.GPIO21, Level::High);
    let dc = Output::new(peripherals.GPIO20, Level::Low);
    let rst = Output::new(peripherals.GPIO19, Level::Low);

    let dma_channel = peripherals.DMA_CH0;
    let dma_buffs = dma_buffers!(2, 32000);
    let rx_buff = DmaRxBuf::new(dma_buffs.1, dma_buffs.0).unwrap();
    let tx_buff = DmaTxBuf::new(dma_buffs.3, dma_buffs.2).unwrap();

    let spi_config = Config::default()
        .with_mode(Mode::_0)
        .with_frequency(80.MHz());
    let spi = Spi::new(peripherals.SPI2, spi_config)
        .unwrap()
        .with_mosi(mosi)
        .with_miso(miso)
        .with_sck(sclk)
        .with_dma(dma_channel)
        .into_async()
        .with_buffers(rx_buff, tx_buff);
    info!("Spi created");

    let spi_bus = Mutex::<NoopRawMutex, _>::new(spi);
    let spi_device = SpiDevice::new(&spi_bus, cs_display);
    log::info!("SpiDevices created");
    let timer0 = esp_hal::timer::systimer::SystemTimer::new(peripherals.SYSTIMER);
    esp_hal_embassy::init(timer0.alarm0);

    info!("Embassy initialized!");

    let mut display_buf = [0u8; 240 * 240 * 2];
    let async_interface = SpiInterfaceAsync::new(spi_device, dc, &mut display_buf, (240, 240));
    log::info!("async DI created");

    let mut display = Builder::new(ST7789, async_interface)
        .reset_pin(rst)
        .display_size(240, 240)
        .orientation(Orientation::new().rotate(Rotation::Deg90))
        .invert_colors(ColorInversion::Inverted)
        .init(&mut delay)
        .unwrap();

    let start = time::now();
    display.clear(Rgb565::BLUE).unwrap();
    display.flush().await.unwrap();
    log::info!("clear drawing time {}", time::now() - start);

    delay.delay_millis(1500);

    let object = Circle::with_center(Point::new(120, 120), 80).into_styled(
        PrimitiveStyleBuilder::new()
            .fill_color(Rgb565::YELLOW)
            .stroke_color(Rgb565::BLACK)
            .stroke_width(2)
            .build(),
    );
   
    let start = time::now();
    object.draw(&mut display).unwrap();
    display.flush().await.unwrap();
    log::info!("object drawing time {}", time::now() - start);

    delay.delay_millis(1500);

    let test_image = TestImage::new();
    let start = time::now();
    test_image.draw(&mut display).unwrap();
    display.flush().await.unwrap();
    log::info!("image drawing time {}", time::now() - start);

    let _ = spawner;
    loop {
        Timer::after(Duration::from_secs(1)).await;
    }
}`
```

rfuest:
I didn't have a lot of time to look at this, so I might not understand every detail yet, but I'm not sure I like this approach of adding async.

Here are some issues that I would like to discuss further:
- `SpiInterfaceAsync` is a `Interface` impl, but behaves differently than all other `Interface` impls by requiring a `flush`.
  This would at least need to be documented.
- Capturing the column address and page address range in `send_command` is clever, but will cause issues.
  The recently added RM67162 reuses the `0x2A` command in a manufacturer specific command set for other purposes: https://github.com/almindor/mipidsi/blob/4e5610d798b328f67c2cb0de2dd0ade7698d242d/src/models/rm67162.rs#L61C13-L61C38. This would cause the code in `SpiInterfaceAsync` to panic. In this case it would be easy to disambiguate this by checking the parameter length, but I don't think it's a good idea to rely on the fact that there will never be another `0x2A` command with 4 parameter bytes.

What do you think about an alternative approach: Instead of trying to mix framebuffer handling and display communication we split it into two separate things. The framebuffer is in memory only and tracks the drawing extents. To draw the framebuffer we could add something like `fn Display::update_from_framebuffer(...)` and `async fn Display::update_from_framebuffer_async(...)`.

almindor:
> I didn't have a lot of time to look at this, so I might not understand every detail yet, but I'm not sure I like this approach of adding async.
> 
> Here are some issues that I would like to discuss further:
> 
> * `SpiInterfaceAsync` is a `Interface` impl, but behaves differently than all other `Interface` impls by requiring a `flush`.
>   This would at least need to be documented.
> * Capturing the column address and page address range in `send_command` is clever, but will cause issues.
>   The recently added RM67162 reuses the `0x2A` command in a manufacturer specific command set for other purposes: https://github.com/almindor/mipidsi/blob/4e5610d798b328f67c2cb0de2dd0ade7698d242d/src/models/rm67162.rs#L61C13-L61C38. This would cause the code in `SpiInterfaceAsync` to panic. In this case it would be easy to disambiguate this by checking the parameter length, but I don't think it's a good idea to rely on the fact that there will never be another `0x2A` command with 4 parameter bytes.
> 
> What do you think about an alternative approach: Instead of trying to mix framebuffer handling and display communication we split it into two separate things. The framebuffer is in memory only and tracks the drawing extents. To draw the framebuffer we could add something like `fn Display::update_from_framebuffer(...)` and `async fn Display::update_from_framebuffer_async(...)`.

Yes, and for full async we would need to introduce a bunch of other things namely:
1. A command stack so `Model::init` calls can remain sync, but be called async (including delay and output pin handling)
2. Optionality of the framebuffer for sync setup, but requirement of it for async (Builder would handle that logic)
3. A "renderer" separation in Display enabling draw calls to be buffered or direct (sync-only for direct)
4. Unification of DI usage into minimal points so we can easily have sync/async separation

Models remain sync, they just pipe commands into a stack instead of executing them directly.
The DI will have to be expanded with a async version, but it will be simpler (no need for repeated pixels, or other optimizations there)
e-g::DrawTarget remains sync, it will either execute direct, or buffer depending on the renderer selected
Direct-Sync versions will simply execute the operations directly
Buffered-Sync version will execute commands directly, but drawing will be deferred to a sync "flush" call
Buffered-Async version will defer commands to a "init_async" and drawing to a async "flush" call

I am unsure if there's a roadbloack, the biggest hurdle is the refactor and the mess with DI (e.g. e-h-a vs e-h SpiDevice etc.) Essentially the glueing of it together.

rfuest:
> Yes, and for full async we would need to introduce a bunch of other things namely:

Making everything async would indeed be a lot of work and I don't think there are enough benefits to justify this at the moment. Having an async way to update the display would be a good MVP in my opinion.

My suggestion was to make the required changes for async support minimal from a `Interface` trait perspective. It would be enough to have a `InterfaceAsync::send_pixels_async(&mut self, region: (u16, u16, u16, u16), pixels: &Framebuffer) -> Future<...>`. This would set up the write by issuing the `SetColumnAddress` and `SetPageAddress` commands and then transfer the pixels. This way there would be no need to buffer any commands or having a separate `flush` command.

almindor:
> > Yes, and for full async we would need to introduce a bunch of other things namely:
> 
> Making everything async would indeed be a lot of work and I don't think there are enough benefits to justify this at the moment. Having an async way to update the display would be a good MVP in my opinion.
> 
> My suggestion was to make the required changes for async support minimal from a `Interface` trait perspective. It would be enough to have a `InterfaceAsync::send_pixels_async(&mut self, region: (u16, u16, u16, u16), pixels: &Framebuffer) -> Future<...>`. This would set up the write by issuing the `SetColumnAddress` and `SetPageAddress` commands and then transfer the pixels. This way there would be no need to buffer any commands or having a separate `flush` command.

That won't work because if you introduce an async interface you can only use it in a fully async setup. This is because `SpiDevice` is either async, or sync. It's never both. You could hack it like I did here with `block_on` but @GrantM11235 explained that that'd just deadlock.

So there's no way to mix and match. We either do async, or sync, never both.

almindor:
Closing in favor of #179

Issue #76, opened 2023-10-06
Link to TROUBLESHOOTING.md in the docs is broken


rfuest:
The link to `TROUBLESHOOTING.md` from `lib.rs` broken, because of the split into the `mipidsi` and `mipidsi-async` directories. We could either move the `docs` directory in the root of the repo or change the links to point to the new location. Moving the `docs` directory into the root would have the advantage that all already released versions of the `mipidsi` docs would be fixed. In the future we might want to consider to update the links to point to a specific tag/release to be resilient against changes in the repo layout.

rfuest:
Instead of fixing the link we could also move the troubleshooting document into the crate docs.

PR #174, opened 2025-02-05
Update TROUBLESHOOTING.md to reflect Builder methods name changes

DivineGod:
Updates to the names of methods on the `Builder` merged in PR #113 missed updating the TROUBLESHOOTING.md file

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -40,7 +40,7 @@ If the test image isn't displayed as expected use one of the reference image bel
 ![Wrong subpixel order](colors_wrong_subpixel_order.svg)
 
 ```rust
-.with_color_order(mipidsi::options::ColorOrder::Bgr)
+.color_order(mipidsi::options::ColorOrder::Bgr)
 ```
 
 ### Wrong color inversion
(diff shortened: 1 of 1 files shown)


rfuest:
Good catch!

There's another part of the document that hasn't been updated in line 26, where `Builder::ili9486_rgb666(di)` instead of the current `Builder::new(ILI9486Rgb666, di)`. Could you also fix that while you are at it?

@almindor: I had previously suggested to move the troubleshooting doc into the rustdoc docs (https://github.com/almindor/mipidsi/issues/76). My primary concern in that issue was that the troubleshooting doc isn't versioned and that older versions on docs.rs won't link to the correct version of the troubleshooting doc. But this issue shows that there would be another benefit of integrating the troubleshooting guide into the rustdoc docs, which is that we could run doc tests to make sure the code examples remain valid. What is your opinion on this?

DivineGod:
> There's another part of the document that hasn't been updated in line 26, where `Builder::ili9486_rgb666(di)` instead of the current `Builder::new(ILI9486Rgb666, di)`. Could you also fix that while you are at it?

Done

rfuest (approved):
Thanks!

PR #179, opened 2025-02-17
implement Async version of mipidsi

almindor:
Supersedes #173 by implementing a `DisplayAsync` via:
1. Refactoring `Model::init` to take a `InitEngine` which then decides what to do with the init commands and delay calls
2. Adding a sync/direct and an async/buffering `InitEngine`
3. Adding a `InterfaceAsync` to complement `Interface` using frame buffering and async flushing
4. Adding `Builder::init_async` to make use of the new async setups
5. Implementing `DisplayAsync` using the new tools
6. Moved `InterfaceExt` into `Interface` directly for simplicity

### DIFF
--- a/src/init_engine.rs
+++ b/src/init_engine.rs
@@ -1,0 +1,136 @@
+use core::marker::PhantomData;
+
+use embedded_hal::delay::DelayNs;
+
+use crate::{
+    dcs::DcsCommand,
+    interface::{Interface, InterfaceAsync, InterfaceKind},
+    models::ModelInitError,
+};
+
+/// todo
+pub trait InitEngine {
+    // required for interface checks in Model::init
+    const INTERFACE_KIND: InterfaceKind;
+    type Error: core::fmt::Debug;
+    /// Queue a raw command with optional parameters
+    fn queue_command_raw(&mut self, command: u8, args: &[u8]) -> Result<(), Self::Error>;
+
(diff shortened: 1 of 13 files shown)


almindor:
@rfuest @GrantM11235 attempt `#2` using queued commands for init handling and separate interface + display implementation for the rest.

Seems to work fine, I do plan to do a bunch of refactoring still but wanted to hear what you think about this kind of approach.

rfuest:
I think that a command queue for the init process is a good idea.

What is your opinion on splitting the framebuffer into another struct? In my opinion the display communication and the framebuffer are separate things. This would also have the benefit of allowing the user to use multiple framebuffers to implement double buffering or for partial display updates using multiple smaller framebuffers.

src/init_engine.rs
```diff
@@ -0,0 +1,136 @@
...
+    delay: &'delay mut DELAY,
+}
+
+pub struct InitEngineAsync<DI> {
+    queue: heapless::Deque<QueuedInitCommand, 24>,
+    _di: PhantomData<DI>,
+}
+
+pub enum QueuedInitCommand {
+    RawDcs(u8, [u8; 16]),
```
rfuest, line 37: This needs an additional parameter for the data length, otherwise all DCS commands will send 16 bytes of parameters. I'm not sure if all controllers handle excess bytes gracefully, but I think it's better not to send them.

okhsunrog:
@almindor is support for embedded-graphics planned?

PR #181, opened 2025-04-09
Add LCD ili9225

darkautism:
First of all, I want to thank the maintainer of `mipidsi`. This library has been incredibly helpful.

I successfully used `mipidsi` to illuminate an LCD based on the ili9225 controller and wrote a new initialization procedure. However, I noticed that the commands for this controller differ significantly from other modules. Unfortunately, these commands are not implemented as traits, so I'm unsure how to handle them effectively. 

For now, I've implemented the changes using conditional compilation with features to avoid impacting the existing codebase. While I admit this approach is not particularly elegant, it's the best solution I could come up with. I’d love to hear your thoughts on this PR and whether you think there's room for improvement.

### DIFF
--- a/src/dcs.rs
+++ b/src/dcs.rs
@@ -82,19 +82,16 @@ dcs_basic_command!(
     ExitSleepMode,
     0x11
 );
-
 dcs_basic_command!(
     /// Enter Partial Mode
     EnterPartialMode,
     0x12
 );
-
 dcs_basic_command!(
     /// Enter Normal Mode
     EnterNormalMode,
     0x13
 );
-
 dcs_basic_command!(
     /// Turn Display Off
     SetDisplayOff,
@@ -106,13 +103,11 @@ dcs_basic_command!(
     SetDisplayOn,
     0x29
 );
-
 dcs_basic_command!(
     /// Exit Idle Mode
     ExitIdleMode,
     0x38
 );
-
 dcs_basic_command!(
     /// Enter Idle Mode
     EnterIdleMode,
@@ -128,7 +123,6 @@ dcs_basic_command!(
 //     EnterInvertMode,
 //     0x20
 // );
-
 dcs_basic_command!(
     /// Initiate Framebuffer Memory Write
     WriteMemoryStart,
--- a/src/dcs.rs
+++ b/src/dcs.rs
@@ -67,72 +67,64 @@ impl<T: Interface> InterfaceExt for T {}
 
 // DCS commands that don't use any parameters
 
-#[cfg(feature = "ili9225")]
 dcs_basic_command!(
     /// Software Reset
-    SoftReset,
+    SoftResetILI9225,
     0x28
 );
 
-#[cfg(feature = "ili9225")]
 dcs_basic_command!(
     /// Software Reset
-    WriteMemoryStart,
+    WriteMemoryStartILI9225,
     0x22
 );
 
-#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Software Reset
     SoftReset,
     0x01
 );
-#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Enter Sleep Mode
     EnterSleepMode,
     0x10
 );
-#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Exit Sleep Mode
     ExitSleepMode,
     0x11
 );
-#[cfg(not(feature = "ili9225"))]
+
 dcs_basic_command!(
     /// Enter Partial Mode
     EnterPartialMode,
     0x12
 );
-#[cfg(not(feature = "ili9225"))]
+
 dcs_basic_command!(
     /// Enter Normal Mode
     EnterNormalMode,
     0x13
 );
-#[cfg(not(feature = "ili9225"))]
+
 dcs_basic_command!(
     /// Turn Display Off
     SetDisplayOff,
     0x28
 );
 
-#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Turn Display On
     SetDisplayOn,
     0x29
 );
 
-#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Exit Idle Mode
     ExitIdleMode,
     0x38
 );
 
-#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Enter Idle Mode
     EnterIdleMode,
@@ -149,7 +141,6 @@ dcs_basic_command!(
 //     0x20
 // );
 
-#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Initiate Framebuffer Memory Write
     WriteMemoryStart,
--- a/src/dcs.rs
+++ b/src/dcs.rs
@@ -3,7 +3,7 @@
 use crate::interface::Interface;
 
 #[macro_use]
-mod macros;
+pub(crate) mod macros;
 
 mod set_address_mode;
 pub use set_address_mode::*;
@@ -66,18 +66,6 @@ pub trait InterfaceExt: Interface {
 impl<T: Interface> InterfaceExt for T {}
 
 // DCS commands that don't use any parameters
-
-dcs_basic_command!(
-    /// Software Reset
-    SoftResetILI9225,
-    0x28
-);
-
-dcs_basic_command!(
-    /// Software Reset
-    WriteMemoryStartILI9225,
-    0x22
-);
 
 dcs_basic_command!(
     /// Software Reset
--- a/src/dcs.rs
+++ b/src/dcs.rs
@@ -67,48 +67,72 @@ impl<T: Interface> InterfaceExt for T {}
 
 // DCS commands that don't use any parameters
 
+#[cfg(feature = "ili9225")]
+dcs_basic_command!(
+    /// Software Reset
+    SoftReset,
+    0x28
+);
+
+#[cfg(feature = "ili9225")]
+dcs_basic_command!(
+    /// Software Reset
+    WriteMemoryStart,
+    0x22
+);
+
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Software Reset
     SoftReset,
     0x01
 );
-
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Enter Sleep Mode
     EnterSleepMode,
     0x10
 );
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Exit Sleep Mode
     ExitSleepMode,
     0x11
 );
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Enter Partial Mode
     EnterPartialMode,
     0x12
 );
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Enter Normal Mode
     EnterNormalMode,
     0x13
 );
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Turn Display Off
     SetDisplayOff,
     0x28
 );
 
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Turn Display On
     SetDisplayOn,
     0x29
 );
+
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Exit Idle Mode
     ExitIdleMode,
     0x38
 );
+
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Enter Idle Mode
     EnterIdleMode,
@@ -124,6 +148,8 @@ dcs_basic_command!(
 //     EnterInvertMode,
 //     0x20
 // );
+
+#[cfg(not(feature = "ili9225"))]
 dcs_basic_command!(
     /// Initiate Framebuffer Memory Write
     WriteMemoryStart,
--- a/src/dcs/set_address_mode.rs
+++ b/src/dcs/set_address_mode.rs
@@ -1,15 +1,58 @@
 //! Module for the MADCTL instruction constructors
 
+use super::DcsCommand;
+use crate::dcs::InterfaceExt;
 use crate::options::{
     ColorOrder, HorizontalRefreshOrder, MemoryMapping, ModelOptions, Orientation, RefreshOrder,
     VerticalRefreshOrder,
 };
 
-use super::DcsCommand;
+/// This is a public trait for user can change their MADCTL command.
+pub trait AddressMode: Copy {
+    /// Returns this Madctl with [ColorOrder] set to new value
+    fn with_color_order(self, color_order: ColorOrder) -> Self;
+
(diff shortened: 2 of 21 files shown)


almindor:
> First of all, I want to thank the maintainer of `mipidsi`. This library has been incredibly helpful.
> 
> I successfully used `mipidsi` to illuminate an LCD based on the ili9225 controller and wrote a new initialization procedure. However, I noticed that the commands for this controller differ significantly from other modules. Unfortunately, these commands are not implemented as traits, so I'm unsure how to handle them effectively.
> 
> For now, I've implemented the changes using conditional compilation with features to avoid impacting the existing codebase. While I admit this approach is not particularly elegant, it's the best solution I could come up with. I’d love to hear your thoughts on this PR and whether you think there's room for improvement.

Interesting, DCS should be universal. Do you have a link to the documentation you used for this? 

If we do need to have differing DCS between models, I think as you said,  traits might be the best approach, or we could hide things in submodules conditionally, similar to your approach but module level to keep things nicer.

darkautism:
According to [[TFT Drivers](https://github.com/Bodmer/TFT_eSPI/tree/master/TFT_Drivers)](https://github.com/Bodmer/TFT_eSPI/tree/master/TFT_Drivers) and [[this specific line in TFT_eSPI.cpp](https://github.com/Bodmer/TFT_eSPI/blob/5793878d24161c1ed23ccb136f8564f332506d53/TFT_eSPI.cpp#L3360)](https://github.com/Bodmer/TFT_eSPI/blob/5793878d24161c1ed23ccb136f8564f332506d53/TFT_eSPI.cpp#L3360), it's clear that the instructions are different, even the instruction widths vary. My implementation is based on these references.

rfuest:
> @rfuest what do you think?

I'm on the fence about adding this. The ILI9225 doesn't follow the DCS command set at all and uses something custom. Adding support for this controller would extend the scope of this crate beyond DCS displays (excerpt from the README: "This crate provides a generic display driver to connect to TFT displays that implement the MIPI Display Command Set."

But on the other hand adding something with a completely different command set would ensure that our code is flexible enough to handle more display types.

If we decide to add this it shouldn't require conditional compilation. The `dcs` module implements a standard command set and it shouldn't be modified by changing a feature. As an alternative it would be possible to add more methods to the `Model` trait, that have default implementations for DCS displays, but can be overridden to support other displays. For example:
```rust
struct Display<...> {
    ...
    fn set_address_window<DI: ...>(&mut self, di: &mut DI, sx: u16, sy: u16, ex: u16, ey: u16) -> Result<(), DI::Error> {
        // add clipping offsets if present
        let mut offset = self.options.display_offset;
        let mapping = MemoryMapping::from(self.options.orientation);
        if mapping.reverse_columns {
            offset.0 = M::FRAMEBUFFER_SIZE.0 - (self.options.display_size.0 + offset.0);
        }
        if mapping.reverse_rows {
            offset.1 = M::FRAMEBUFFER_SIZE.1 - (self.options.display_size.1 + offset.1);
        }
        if mapping.swap_rows_and_columns {
            offset = (offset.1, offset.0);
        }

        let (sx, sy, ex, ey) = (sx + offset.0, sy + offset.1, ex + offset.0, ey + offset.1);

        MODEL::update_address_window(&mut self.di, sx, sy, se, ey)
    }
    ...
}

trait Model {
    ...
    // Default impl for DCS compliant models.
    fn update_address_window<DI: ...>(di: &mut DI, sx: u16, sy: u16, ex: u16, ey: u16) -> Result<(), DI::Error> {
        di.write_command(dcs::SetColumnAddress::new(sx, ex))?;
        di.write_command(dcs::SetPageAddress::new(sy, ey))
    }
    ...
}

impl Model for ILI9225Rgb565 {
    ...
    fn update_address_window<DI: ...>(di: &mut DI, sx: u16, sy: u16, ex: u16, ey: u16) -> Result<(), DI::Error> {
        di.write_raw(0x37, &sx.to_be_bytes())?;
        di.write_raw(0x36, &ex.to_be_bytes())?;
        di.write_raw(0x39, &sy.to_be_bytes())?;
        di.write_raw(0x38, &ey.to_be_bytes())?;
        di.write_raw(0x20, &sx.to_be_bytes())?;
        di.write_raw(0x21, &sy.to_be_bytes())
    }
    ...
}
```

darkautism:
I remove feature flag and change other LCD using default trait.

WIP: Add rotation feature

darkautism:
Finished.

I Add new trait to resolve different size of madctl.

darkautism:
I also notice ```.orientation(Orientation::new().rotate(Rotation::Deg270))``` not been called when we using Builder. Is this the expected behavior?

I must rerun 
```Rust
    display
        .set_orientation(Orientation::default().rotate(config::ROTATION))
        .unwrap();
        ```

File: src/dcs/set_address_mode.rs [hunk 1]
```diff
@@ -1,16 +1,59 @@
...
+use super::DcsCommand;
+use crate::dcs::InterfaceExt;
 use crate::options::{
     ColorOrder, HorizontalRefreshOrder, MemoryMapping, ModelOptions, Orientation, RefreshOrder,
     VerticalRefreshOrder,
 };
 
-use super::DcsCommand;
+/// This is a public trait for user can change their MADCTL command.
+pub trait AddressMode: Copy {
```
rfuest, line 11: This trait adds a lot of additional trait bounds and I don't think we need it. See my comment for `Display::madctl` below for a possible alternative.

src/lib.rs
```diff
@@ -153,7 +154,7 @@ where
     // Model Options, includes current orientation
     options: options::ModelOptions,
     // Current MADCTL value copy for runtime updates
-    madctl: dcs::SetAddressMode,
+    madctl: MODEL::AddressMode,
```
rfuest, line 157: Instead of the associated `Model::AddressMode` type, I would suggest to store the `ModelOptions` in this struct and add a `update_options(&mut DI, options: &ModelOptions) -> Result<(), DI::Error>` method to the `Model` trait.

src/lib.rs
```diff
@@ -355,7 +355,7 @@ where
     /// Need to call [Self::wake] before issuing other commands
     ///
     pub fn sleep<D: DelayNs>(&mut self, delay: &mut D) -> Result<(), DI::Error> {
-        self.di.write_command(dcs::EnterSleepMode)?;
+        M::sleep(&mut self.di, delay)?;
         // All supported models requires a 120ms delay before issuing other commands
         delay.delay_us(120_000);
```
rfuest, line 360: I think this delay should be moved into the implementations of this method. `Model::sleep` and `Model::wake` do already get passed the `DelayNs` anyway and it would be more flexible for implementations to be able to override this delay if a controller supports faster sleep/wake.

rfuest (changes requested):
Looks much better than the previous version.

src/lib.rs
darkautism: 圖片
According to the example code in the datasheet, it writes a command with a delay, so I passed it into a function. For other LCDs, it’s just one command, so I kept the code in its original place. Should I move the delay into the wake/sleep function instead?

src/dcs.rs
```diff
@@ -67,12 +67,23 @@ impl<T: Interface> InterfaceExt for T {}
...
+    SoftResetILI9225,
+    0x28
+);
+
+dcs_basic_command!(
+    /// Software Reset
+    WriteMemoryStartILI9225,
+    0x22
+);
+
```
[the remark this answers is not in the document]
darkautism: Moved

src/lib.rs
darkautism: I will fix it tomorrow.

darkautism: It seems that madctl will never be used again.

src/lib.rs
rfuest: Yes, moving the delay into `Model` is more flexible than having the fixed delay in `Display::{sleep, wake}`. The methods in `Display` should only call the corresponding method in `Model` and set the `sleeping` flag.

rfuest:
I noticed one more issue while I looked at the application note for the ILI9225. The reference code drives the reset pin low for 10 ms and contains a comment that this delay is necessary. According to the datasheet the reset pulse needs to be at least 1 ms long.

This timing requirement is violated by our reset implementation, which only drives reset low for 10 µs:
https://github.com/almindor/mipidsi/blob/4e5610d798b328f67c2cb0de2dd0ade7698d242d/src/builder.rs#L179-L181

I would suggest to add a constant for the minimum reset time to the `Model` trait to allow `Model` impls to specify a longer reset duration.

src/models/ili9225.rs
```diff
@@ -0,0 +1,251 @@
...
+    let rotation = (options.orientation.rotation as u8) % 4; // Only accept 0-3
+
+    // Command 1: DRIVER_OUTPUT_CTRL (0x01)
+    let driver_high_byte = match rotation {
+        0 => 0x01, // 0°
+        1 => 0x00, // 90°
+        2 => 0x02, // 180°
+        3 => 0x03, // 270°
+        _ => 0x01, // Not reachable
+    };
```
rfuest, line 69: Why are you converting the enum to an `u8` and the match the value instead of matching rotation directly?
```suggestion
    // Command 1: DRIVER_OUTPUT_CTRL (0x01)
    let driver_high_byte = match options.orientation.rotation {
        Rotation::Deg0 => 0x01,
        Rotation::Deg90 => 0x00,
        Rotation::Deg180 => 0x02,
        Rotation::Deg270 => 0x03,
    };
```

src/models/ili9225.rs
```diff
@@ -0,0 +1,251 @@
...
+    let color_order_byte = if options.color_order == ColorOrder::Bgr {
+        0x10
+    } else {
+        0x00
+    };
+    let entry_low_byte = if rotation == 1 || rotation == 3 {
+        0x38
+    } else {
+        0x30
+    };
```
rfuest, line 83:
```suggestion
    let entry_low_byte = if options.orientation.rotation.is_vertical() {
        0x38
    } else {
        0x30
    };
```

src/models/ili9225.rs
```diff
@@ -0,0 +1,251 @@
...
+    let driver_params = [driver_high_byte, 0x1C];
+    di.write_raw(ILI9225_DRIVER_OUTPUT_CTRL, &driver_params)?;
+
+    // Command 2: ENTRY_MODE (0x03)
+    let color_order_byte = if options.color_order == ColorOrder::Bgr {
+        0x10
+    } else {
+        0x00
+    };
```
rfuest, line 78: Writing this using a match statement is a bit shorter and safer if we ever decide to add another color order.
```suggestion
    let color_order_byte = match options.color_order {
        ColorOrder::Rgb => 0x00,
        ColorOrder::Bgr => 0x10,
    };
```

rfuest (changes requested):
Some more things I notices:
- `set_tearing_effect`, `set_vertical_scroll_region`, and `set_vertical_scroll_offset` are still hardcoded to use DCS commands and would fail on the ILI9225. They should either be supported or disabled.
- The init routine doesn't use the `color_inversion` setting.

src/lib.rs
rfuest: I hadn't noticed that we already store the `ModelOptions` in `Display`.

> It seems that madctl will never be used again.

Good catch, if that is the case you could just remove it.

src/models/ili9225.rs
darkautism: Fixed

src/lib.rs
darkautism: Some test case use this field so i not remove it.

src/lib.rs
```diff
@@ -395,9 +397,9 @@ pub mod _mock {
 
     use embedded_hal::{delay::DelayNs, digital, spi};
 
+    use crate::models::ILI9341Rgb565;
```
rfuest, line 400: Please revert this unnecessary change.

src/models.rs
```diff
@@ -37,6 +44,9 @@ pub trait Model {
     /// The framebuffer size in pixels.
     const FRAMEBUFFER_SIZE: (u16, u16);
 
+    /// Delay after lcd reset.
+    const LCD_RESET_DELAY: u32 = 10;
```
rfuest, line 48: Let's use something a bit more descriptive for this:
```suggestion
    /// Duration of the active low reset pulse in µs.
    const RESET_DURATION: u32 = 10;
```

src/models/ili9225.rs
```diff
@@ -0,0 +1,250 @@
...
+        options_write_cmd(di, options)
+    }
+}
+
+#[path = "../dcs/macros.rs"]
+#[macro_use]
+mod dcs_macros;
+
+dcs_basic_command!(
+    /// Software Reset
```
rfuest, line 241: Copy and paste error:
```suggestion
    /// Initiate Framebuffer Memory Write
```

rfuest (changes requested):

src/models/ili9225.rs
darkautism: Fixed

src/models.rs
darkautism: Fixed

src/lib.rs
```diff
@@ -366,7 +368,7 @@ where
     /// Wakes the display after it's been set to sleep via [Self::sleep]
     ///
     pub fn wake<D: DelayNs>(&mut self, delay: &mut D) -> Result<(), DI::Error> {
-        self.di.write_command(dcs::ExitSleepMode)?;
+        M::wake(&mut self.di, delay)?;
         // ST7789 and st7735s have the highest minimal delay of 120ms
         delay.delay_us(120_000);
```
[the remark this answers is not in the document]
darkautism: Fixed

almindor (changes requested):
I think I'll be happy with this once CHANGELOG entry is added.

Pushing some of the lib commands into Model is a bit questionable but it gives us support for these custom displays and I think that's acceptable (I just won't be happy about it :D)

rfuest:
> Pushing some of the lib commands into Model is a bit questionable but it gives us support for these custom displays and I think that's acceptable (I just won't be happy about it :D)

What exactly makes you unhappy about this? Would you prefer that the additional methods wouldn't exist at all or do you dislike having default implementations in `Model`?

darkautism:
color_inversion not used in TFT_eSPI:
https://github.com/Bodmer/TFT_eSPI/blob/5793878d24161c1ed23ccb136f8564f332506d53/TFT_Drivers/ILI9225_Defines.h#L38
But it seems in datasheet. Try to impl by myself.

darkautism:
Is this behavior is correct?

rfuest:
> Is this behavior correct?

That looks correct. You can also use the `TestImage` to check and refer to this chart: https://github.com/almindor/mipidsi/blob/v0.9.0/docs/TROUBLESHOOTING.md#incorrect-colors

Your code does currently set the `REV` bit in the "Display Control 1" register when the color inversion is set to `Normal`, which is the wrong way around. This might be counter intuitive, because it works correct on your display, but we decided to make this consistent across all models and require users of IPS panels to add `.with_invert_colors(ColorInversion::Inverted)` to their builder call.

almindor (approved):
LGTM

rfuest (approved):
Some of the docs could be improved a bit, but I'll create a follow up PR for that. The rest looks good and can be merged. Thanks for the PR.

PR #185, opened 2025-09-07
Adding defmt::Format support to interface errors

pebender:
Thank you for mipidsi.

Would it be possible to add feature gated defmt::Format support to the interface errors?

This pull request adds defmt::Format support to the SPI and parallel interface errors. The support is gated behind the feature "defmt". The feature enables "defmt" support in mipidsi. In addition, it enables embedded-hal's defmt support so that embedded-hal's ErrorKind error will be sure to have defmt::Format support. It does not enable embedded-graphics-core's defmt support because the SPI and parallel interface errors do not appear to have any dependency on embedded-graphics-core.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mipidsi"
 description = "MIPI Display Command Set compatible generic driver"
-version = "0.9.0"
+version = "0.9.1"
 authors = ["Ales Katona <almindor@gmail.com>"]
 edition = "2021"
 license = "MIT"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "mipidsi"
 description = "MIPI Display Command Set compatible generic driver"
-version = "0.9.0"
```
almindor, line 4: Could you unbump this? We'll do the version bump as releases happen. Also please add a CHANGELOG entry for this change.

almindor (changes requested):
LGTM, just needs a CHANGELOG entry and removal of the version bump.

Thanks!

pebender:
Oops, I closed this by deleting my fork. I will do a new pull request with the requested changes.

PR #186, opened 2025-09-09
added defmt support for Interface errors

pebender:
Because I accidentally closed #185 by deleting my fork. Here is a new pull request with the requested changes.

Again, thank you for mipidsi.

This pull request adds defmt::Format support to the SPI and parallel interface errors. The support is gated behind the feature "defmt". The feature enables "defmt" support in mipidsi. In addition, it enables embedded-hal's defmt support so that embedded-hal's ErrorKind error will be sure to have defmt::Format support. It does not enable embedded-graphics-core's defmt support because the SPI and parallel interface errors do not appear to have any dependency on embedded-graphics-core.


almindor (approved):
LGTM, thank you!

PR #188, opened 2025-10-11
Add defmt::Format derives whenever there is a Debug derive.

pebender:
As little while ago I summited a pull request to add defmt support to interface errors. As I have improved the error handling in my project, I have found that I needed more structs/enums to have defmt support. For example, I needed defmt support for InitError. So, I added defmt support wherever there was Debug support in hopes that would catch everything.


almindor (approved):

pebender:
Thanks.

Issue #193, opened 2026-02-10
Is the ili9488 supported?


rafaeljacov:
it is listed in the README but in the docs it does not have type for it

rfuest:
Support for the ILI9488 has been added in the development version, but the release on crates.io doesn't include it yet.

The ILI9488 is very similar to the ILI9486, which is supported in the released version. You should be able to use that model instead until the next release of the crate.

rafaeljacov:
Thanks! Is there an expected release date for the next release candidate?

almindor:
I didn't even realize we had something cooking up in master. @rfuest should we do a release?

rfuest:
@almindor: And I didn't realize it has been over a year since the last release. We should do a release.

rafaeljacov:
I hope I can help but I'm still a noob and need more guidance on the code architecture

almindor:
v0.10.0 was just released an is now available in crates.io this should fix the issue

PR #194, opened 2026-02-10
release v0.10.0

almindor:
@rfuest lmk if I missed something, once we merge I'll push to crates.io

Solves #193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.10.0] - 2026-02-17
+
 ### Added
 
 - added `SpiInterface::release`
@@ -30,7 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - remove unused `nb` dependency
 - remove panic from `Builder::display_size` if invalid size is given
 
-## [0.9.0]
+## [v0.9.0] - 2025-01-10
 
 ### Added
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mipidsi"
 description = "MIPI Display Command Set compatible generic driver"
-version = "0.9.0"
+version = "0.10.0"
 authors = ["Ales Katona <almindor@gmail.com>"]
 edition = "2021"
 license = "MIT"
@@ -13,7 +13,7 @@ rust-version = "1.75"
 
 [dependencies]
 defmt = { version = "1.0.1", optional = true }
-embedded-graphics-core = "0.4.0"
+embedded-graphics-core = "0.4.1"
 embedded-hal = "1.0.0"
 
 [dependencies.heapless]
@@ -21,7 +21,7 @@ optional = true
 version = "0.8.0"
 
 [dev-dependencies]
-embedded-graphics = "0.8.1"
+embedded-graphics = "0.8.2"
 
 [features]
 default = ["batch"]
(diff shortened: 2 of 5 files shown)


almindor:
Hmm seems people have broken edition support with dependencies of dependencies. I'll have to look, as I don't want to force migration to a new MSRV yet.

rfuest:
I did notice the dependency issue in e-g recently but didn't have time to look into it. Let me know if you find a solution or if I need to create a new patch release of e-g 0.8 that fixes the `az` dependency issue by limiting it to an older version or removing it entirely.

almindor:
> I did notice the dependency issue in e-g recently but didn't have time to look into it. Let me know if you find a solution or if I need to create a new patch release of e-g 0.8 that fixes the `az` dependency issue by limiting it to an older version or removing it entirely.

yeah i think that'd help, i'm a bit short on time atm. but will try to get back to this end of week

rfuest:
I've created a [new branch ](https://github.com/rfuest/embedded-graphics/tree/eg-0.8)for the maintenance release of e-g 0.8 to fix the dependency issue. Can you try by temporarily using a GIT dependency before I publish 0.8.2?

almindor:
> I've created a [new branch ](https://github.com/rfuest/embedded-graphics/tree/eg-0.8)for the maintenance release of e-g 0.8 to fix the dependency issue. Can you try by temporarily using a GIT dependency before I publish 0.8.2?

I'm guessing latest e-g master is not compatible with mipidsi? I don't have time to look into this atm. will check back later.

rfuest:
I think the error is caused by the `dev-dependency` to `embedded-graphics` which also needs to be changed to the GIT dependency.

almindor:
> I think the error is caused by the `dev-dependency` to `embedded-graphics` which also needs to be changed to the GIT dependency.

That fixed most but somehow not all, I'll try to look  into this more later this weekend/early next week.

rfuest:
The examples have their own Cargo.toml and still depend on the crates.io version of `embedded-graphics`. Everything should work after e-g is updated and I don't think you need to do anything in this crate until I've released the new e-g version.

I've created a [PR](https://github.com/embedded-graphics/embedded-graphics/pull/806) for the dependency fix, but apparently it still needs some work because CI is unhappy.

almindor:
@rfuest so this passes CI with your ref everywhere now. I think I'll see if we can refactor things into a workspace so we keep all dependencies in one cargo.toml file while you sort out the e-g release.

rfuest:
Can you try it again with the latest commit https://github.com/rfuest/embedded-graphics/commit/77372e40ff6b06c779c152edf9cb2b68d79fddb9?

> @rfuest so this passes CI with your ref everywhere now. I think I'll see if we can refactor things into a workspace so we keep all dependencies in one cargo.toml file while you sort out the e-g release.

I believe we intentionally left the examples separate because they use different target architectures.

rfuest:
e-g-core 0.4.1 and e-g 0.8.2 are now released.

almindor:
@rfuest should be good now

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.10.0]
```
rfuest: The entries before 0.9.0 contain the release date, should we also include the date for 0.9 and 0.10?

rfuest (approved):
LGTM, I'm glad the dependency weirdness is fixed, because it was a bit of a pain to get it working in e-g.

If you want you could add the missing dates in the changelog for 0.9 and 0.10 before the release, but it's not a release blocker IMO.

PR #199, opened 2026-04-04
Support Bgr565 as used by the Cheap Yellow Display

deets:
Not much to say about this one, the CYD is pretty popular so this would be nice to upstream. Tested of course (and that's the reason I didn't copy & past implement other Bgr variants just to be safe).

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -1,7 +1,7 @@
 //! Interface traits and implementations
 
 mod spi;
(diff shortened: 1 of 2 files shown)


rfuest:
What is the advantage of this over setting `ColorOrder::Bgr`?

deets:
Nothing. I wasn't aware that this solves the problem as well 🤦 Just tested it, works of course. Thanks, and sorry for the bother.

PR #201, opened 2026-05-08
Add get_display_width and get_display_height

### LINKED ISSUE #200: [Enhancement] add getter for width and height
Add getters `Display::get_width(&self)` and `Display::get_height(&self)` for the library. It would be a lot better if I don't have to manually manage the display size with extra variables and traits

**To Reproduce**
Add the getters `Display::get_width(&self)` and `Display::get_height(&self)` or `Display::get_size(&self) -> (u16, u16)`

**Expected behavior**
`Display.get_width()`, `Display.get_height()`

**Used hardware and software**
Please provide details about the hardware and software you are using:
- `mipidsi` version:  `0.10.0`
- MCU: ESP32
- HAL: e.g. esp_hal 1.1.1
- Display Controller IC: e.g. ST7789v
- Type of connection: SPI

chunix64:
Fix #200

You should consider renaming the methods to `get_width` and `get_height` for shorter names or keep it clear

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -169,15 +169,21 @@ where
     ///
     /// Get display width
     ///
-    pub fn get_display_width(&self) -> u16 {
-        self.options.display_size.0
+    pub fn get_width(&self) -> u16 {
+        match self.options.orientation.rotation {
+            options::Rotation::Deg90 | options::Rotation::Deg270 => self.options.display_size.1,
+            _ => self.options.display_size().0,
+        }
     }
 
     ///
     /// Get display height
     ///
-    pub fn get_display_height(&self) -> u16 {
-        self.options.display_size.1
+    pub fn get_height(&self) -> u16 {
+        match self.options.orientation.rotation {
+            options::Rotation::Deg90 | options::Rotation::Deg270 => self.options.display_size.0,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -166,6 +166,20 @@ where
     M::ColorFormat: InterfacePixelFormat<DI::Word>,
     RST: OutputPin,
 {
+    ///
+    /// Get display width
+    ///
+    pub fn get_display_width(&self) -> u16 {
+        self.options.display_size.0
```
almindor: this is incorrect, depending on what you mean by `display_size`

See options.rs and how we return the `display_size()` incorporating orientation.

I think a read-only getter for `self.options` might be the best way to tackle this, so that all of them can be read back.

almindor (changes requested):

src/lib.rs
rfuest: > I think a read-only getter for `self.options` might be the best way to tackle this, so that all of them can be read back.

If we do this we need to decide if we want to expose both the `display_size` field and the `display_size()` function, because having both with the same name but different semantics is confusing. Perhaps we can find a better way to distinguish between default orientation display size and the display size with orientation applied by naming them differently.

PR #202, opened 2026-05-10
Add support for NV3007

samphonic:
This change adds support for the NV3007 model (standard variant, not 279 variant).

The initialization sequence was based on code found here:
- https://community.home-assistant.io/t/use-display-with-nv3007-controller/936485
- https://github.com/variegated-coffee/variegated-rs/blob/main/variegated-nv3007/src/displays/nv3007.rs
- https://github.com/moononournation/Arduino_GFX/blob/master/src/display/Arduino_NV3007.h

I have done some light testing and it definitely works in basic circumstances (using async SPI and DMA), but might be missing some features (I haven't attempted to implement any inverted options, for example). I tried to re-use existing commands where possible, but since a lot of the initialization was just lifted from other libraries I may have missed something.

There is a line in the initial sequence labeled "Pixel format" but the usual `SetPixelFormat` command was also required at the end-- unclear what that first one is for.


src/models/nv3007.rs
```diff
@@ -0,0 +1,196 @@
+use embedded_graphics_core::pixelcolor::Rgb565;
+use embedded_hal::delay::DelayNs;
+
+use crate::{
+    ConfigurationError, dcs::{
+        BitsPerPixel, ExitSleepMode, InterfaceExt, PixelFormat, SetAddressMode, SetDisplayOn, SetPixelFormat
+    }, interface::{Interface, InterfaceKind}, models::{Model, ModelInitError}, options::ModelOptions
+};
+
+/// ST7796 display in Rgb565 color mode.
```
rfuest:
```suggestion
/// NV3007 display in Rgb565 color mode.
```

src/models/nv3007.rs
```diff
@@ -0,0 +1,196 @@
...
+        delay: &mut DELAY,
+        options: &ModelOptions,
+    ) -> Result<SetAddressMode, ModelInitError<DI::Error>>
+    where
+        DELAY: DelayNs,
+        DI: Interface,
+    {
+        if !matches!(
+            DI::KIND,
+            InterfaceKind::Serial4Line | InterfaceKind::Parallel8Bit | InterfaceKind::Parallel16Bit
```
rfuest: The controller doesn't support a parallel interface.
```suggestion
            InterfaceKind::Serial4Line
```

src/models/nv3007.rs
```diff
@@ -0,0 +1,196 @@
...
+        di.write_raw(0xEC, &[0x04])?;  // repeated per original
+        
+
+        di.write_raw(0xFF, &[0x00])?; // Disable private registers
+
+        // Tearing effect and other settings
+        di.write_raw(0x35, &[0x00])?;
+        di.write_raw(0x44, &[0x00, 0x10])?;
+        di.write_raw(0x46, &[0x10])?;
+
```
rfuest: This enables the TE pin, but `Model`s should default to TE off. The TE pin can be activated after init by using `Display::set_tearing_effect`, if necessary.

The 0x44 and 0x46 are related to an extended TE mode, which we don't support. Keeping the default values should make TE behave like on any other model.

```suggestion
```

src/models/nv3007.rs
```diff
@@ -0,0 +1,196 @@
...
+        //
+        //// END INITIALIZATION BLOCK
+        // 
+
+        // Set orientation, etc.
+        di.write_command(madctl)?;
```
rfuest, line 179: You mentioned that you didn't implement color invert. The controller supports the default commands for inversion. This should be all that is required to add support:
```suggestion
        di.write_command(madctl)?;
        di.write_command(SetInvertMode::new(options.invert_colors))?;
```

rfuest (changes requested):
Thanks for the PR. Looks good except for a few nits. Please also add an entry to the changelog.

samphonic:
Thanks for the quick feedback, I'll try to get this updated tonight.

samphonic:
Updated.

After removing the TE lines, it seems like the display updates differently (I swear it used to have a vertical wipe visible, but now it's sort of diagonal). I tried using `set_tearing_efffect` but none of the settings seemed to make a difference. I tried putting the original 3 lines back and... it's still the same. So I don't know if I totally imagined the original vertical wipe or what, but it's working fine without those lines and I can't tell that they do anything 🤷‍♂️. Maybe if I ever find better documentation somewhere I'll come back with an update.

rfuest:
> After removing the TE lines, it seems like the display updates differently (I swear it used to have a vertical wipe visible, but now it's sort of diagonal). I tried using `set_tearing_efffect` but none of the settings seemed to make a difference. I tried putting the original 3 lines back and... it's still the same. So I don't know if I totally imagined the original vertical wipe or what, but it's working fine without those lines and I can't tell that they do anything 🤷‍♂️. Maybe if I ever find better documentation somewhere I'll come back with an update.

That's strange. TE should only be an output and the settings not influence the output if you don't sync on the TE signal in your application.

rfuest (changes requested):
Looks good. I'll merge it after you've fixed the formatting issue CI complains about and added a changelog entry.

samphonic:
Oops, forgot about the changelog (and the `fmt` issue was in some code I didn't modify, so I wasn't sure-- fixed now though)!

rfuest (approved):
LGTM.

